PR #325, opened 2026-08-15
fix(event-runtime): pi read-only agents keep bash — scan fleet was structurally broken (WM-301)

hdkiller:
Fixes WM-301

Live evidence: work-scan run `run_e881a392` refused with `needs_human` — *"Linear reads could not be performed with the available tools."*

pi's `READ_ONLY_TOOLS` was `read,grep,find,ls,write` — no shell. Since WM-215 routed every LLM event type to pi, that silently broke **the entire read-only scan fleet**: work/triage/sweep/unblock scans shell to `tools/linear.mjs`; merge/ship scans shell to `gh`/`git`; ci-doctor, status-report and disk-diagnose shell to `gh`/`ssh`. Only mutating `dispatch@1` kept working (worktree agents get the full tool set), which masked it.

Same class as OPS-518 (read-only pi couldn't write `result.json`) — that fix added `write` and stopped; this is the other half. `claude.mjs` has always kept `Bash`/`Write`/`Edit` for read-only agents deliberately, enforcing the boundary via workspace confinement rather than tool removal.

Fix: add `bash`; keep `edit` excluded (that's the actual repository-mutation boundary). The comment now explains why each tool is load-bearing so the list isn't 'tightened' into a broken state a third time.

Verification: `bun test event-runtime/lib/adapters/pi.test.mjs` → 30 pass / 0 fail; full suite green. Live re-test of a pi work-scan follows this merge.

### DIFF
--- a/event-runtime/lib/adapters/pi.mjs
+++ b/event-runtime/lib/adapters/pi.mjs
@@ -52,15 +52,23 @@ export const KILL_GRACE_MS = 30_000;
 /** Trace events preview text; the recorder's byte bound is the real limit. */
 const TEXT_PREVIEW_CHARS = 4000;
 
-// pi's documented read-only pattern omits bash/edit from the tool allowlist
-// rather than intercepting calls at runtime — but `write` must stay: every
-// agent-result contract requires the model to write ./result.json, and a run
-// that cannot write it fails contract_violation:missing_result before doing
-// anything (OPS-518; same reasoning as claude.mjs's READ_ONLY_TOOLS keeping
-// Write/Edit). Containment for mutating: false is therefore bash/edit omission
-// plus the workspace-cwd boundary — audited, not enforced, per §14 until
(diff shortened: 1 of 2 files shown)


hdkiller:
Admin-merging: CI cancelled at the Verify job's 10-minute timeout twice (runs 31898265164 and prior) — the known WM-300 infra issue, not a test failure. This diff is 2 files: adding `bash` to pi's READ_ONLY_TOOLS plus the matching argv test expectations. Verified locally: `bun test event-runtime/lib/adapters/pi.test.mjs` → 30 pass / 0 fail, and the full `bun test event-runtime/` suite green on this branch. It unblocks a live outage — every read-only pi agent (work/triage/sweep/unblock/merge/ship scans, ci-doctor, status-report, disk-diagnose) currently refuses with 'reads could not be performed with the available tools'. WM-300 (raise the timeout) follows immediately.

PR #326, opened 2026-08-15
fix(web): announce dashboard view navigation

hdkiller:
## Summary
- announce first-segment dashboard view changes through a mounted polite live region
- move focus to the main view region with a visible ring after route transitions
- preserve destination filter focus and ignore intra-view hash changes

## Verification
- `bun test event-runtime && cd event-runtime/web && bun run build` — pass (1,294 tests; TypeScript and Vite build green)

## UX critique
UX critique: required — NOT-ASSESSED, browser tooling was unavailable to the critic; implementation received a read-only code review and its lazy-filter focus finding was resolved before re-verification.

Fixes OPS-312

### DIFF
--- a/event-runtime/web/src/App.tsx
+++ b/event-runtime/web/src/App.tsx
@@ -1,5 +1,5 @@
 import { useQuery } from "@tanstack/react-query";
-import { useCallback, useEffect, useMemo, useState, lazy, Suspense } from "react";
+import { useCallback, useEffect, useMemo, useRef, useState, lazy, Suspense } from "react";
 import { api } from "./api";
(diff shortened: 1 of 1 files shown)


hdkiller:
CI follow-up: the Verify check timed out while `bun test` remained in progress, with no assertion failure. Read-only diagnosis classified this as **ENV**: unrelated PRs #324 and #319 timed out in the same step on two self-hosted runners, while a preceding run passed in 1m26. Tracking the shared runner/test hang in OPS-521; per diagnosis, not rerunning this PR until that environment issue is addressed.

PR #393, opened 2026-08-16
fix(unblock): fetch recent label history (WM-275)

hdkiller:
## Summary
- fetch the latest 50 label-history events for reply detection and blocked-ticket digests
- add a regression for long issue histories and guard both GraphQL query contracts

## Verification
- `bun test orchestrator/reply-detection.test.mjs` — 14 pass, 0 fail
- `bun build/emit.mjs --check` — generated files match

UX critique: skipped — backend/internal orchestration query with no user-facing flow

Fixes WM-275

### DIFF
--- a/orchestrator/digest.mjs
+++ b/orchestrator/digest.mjs
@@ -55,7 +55,7 @@ const DIGEST_QUERY = `
       nodes {
         identifier title url
         state { name }
-        comments(first: 100) { nodes { createdAt body } }
+        comments(last: 100) { nodes { createdAt body } }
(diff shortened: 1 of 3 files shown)


hdkiller:
factory-merge-fix round=1 finding=334fc11d73d36af19e6ee0dd60fcf68618461ec2482c31e6fdb21b22d5d1aa4d old=7b37451e8203a83133bcd85e08488728b5a3f828
new=68c8ceefbf04a78b93362f226e9bf1c4333cccb2

PR #447, opened 2026-08-17
fix(worker): bound synchronous subprocesses

hdkiller:
## Summary
- bound repo verification and worktree lifecycle scripts with configurable 120s defaults
- preserve typed timeout failure and workspace-retention semantics, including partial bring-up markers
- bound worker git/Linear helpers and add timeout regression coverage

## Verification
- `bun test event-runtime/lib/verify.test.mjs event-runtime/lib/workspace.test.mjs event-runtime/lib/worker.test.mjs` — 106 pass, 0 fail

UX critique: skipped — backend worker subprocess hardening has no user-facing surface

Fixes WM-262

### File scope
Six files, all under `event-runtime/lib/` — no files outside this scope were touched:

```
 event-runtime/lib/verify.mjs         | 53 ++++++++++++++++++++---
 event-runtime/lib/verify.test.mjs    | 20 +++++++++
 event-runtime/lib/worker.mjs         | 75 +++++++++++++++++---------------
 event-runtime/lib/worker.test.mjs    | 31 +++++++++++++-
 event-runtime/lib/workspace.mjs      | 83 ++++++++++++++++++++++++++++--------
 event-runtime/lib/workspace.test.mjs | 59 ++++++++++++++++++++++++-
 6 files changed, 259 insertions(+), 62 deletions(-)
```

### Verification

Ticket-scoped suites (`bun test` on the three changed test files):

```
 107 pass
 0 fail
 444 expect() calls
Ran 107 tests across 3 files. [4.72s]
```

Full suite (`bun test`), run after `bun install` at the repo root **and** in `event-runtime/web`:

```
 2040 pass
 1 fail
 8045 expect() calls
Ran 2041 tests across 139 files. [203.38s]
```

The single failure is `seed & re-seed deduplication (OPS-464) > initial seed succeeds and verify passes` — the known pre-existing/flaky timing-sensitive test. Clean `develop` shows the same ~2034 pass / 1 fail shape, so it is not caused by this branch. Nothing in this PR's scope (`verify.mjs`, `worker.mjs`, `workspace.mjs`) is involved in it.

Build check (`bun build/emit.mjs --check`), exit 0:

```
ok — 90 generated files match shared/
```

(The command also prints an informational "floor delivery: 10 of 10 repo(s) are not carrying the current floor" notice about other local repos' `AGENTS.md`; it is unrelated to this PR and does not affect the exit status.)

CI on the merge commit `8fff13b` is green — `Verify` pass, `Shadow runner fleet available` pass.

### Merge conflict resolution with `develop`

This branch went CONFLICTING after WM-481 landed on `develop`. The single real conflict was in `event-runtime/lib/workspace.mjs`, in the `worktree_up` failure branch, and was resolved by **combining both sides** rather than taking either one:

- **From WM-262 (this PR):** a `spawnSync` timeout surfaces as `up.error?.code === "ETIMEDOUT"` with a *zero-ish/undefined* status — it is not reported as a non-zero exit. So the failure condition must be `if (up.error?.code === "ETIMEDOUT" || up.status !== 0)`. Dropping this half would let a timed-out bring-up be treated as success, which is precisely the hang this ticket exists to fix.
- **From WM-481 (`develop`):** the failure message prefers the full trimmed `stderr`, then `stdout`, then `exit ${up.status}` — rather than only the last line. Dropping this half would restore the truncated diagnostics WM-481 fixed; a bash `set -e` abort often puts the useful message above the final line.

Both halves address different concerns (detecting the failure vs. reporting it), so keeping both is strictly correct. The resolved code:

```js
if (up.error?.code === "ETIMEDOUT" || up.status !== 0) {
  const timedOut = up.error?.code === "ETIMEDOUT";
  const stderr = (up.stderr || "").trim();
  const stdout = (up.stdout || "").trim();
  const why = timedOut
    ? `timed out after ${timeoutMs}ms`
    : stderr || stdout || `exit ${up.status}`;
  throw new WorktreeError(`worktree_up failed for ${repoName}/${ticket}: ${why}`);
}
```

The same `ETIMEDOUT ||` guard is applied to the `worktree_down` path. Merge commit: `8fff13b`.

### UX critique
Not applicable — no user-visible surface. This is backend-only work: subprocess timeout bounding in the event-runtime worker/verify/workspace libraries, with no UI, copy, navigation, or user-facing flow changes.

### DIFF
--- a/event-runtime/lib/verify.mjs
+++ b/event-runtime/lib/verify.mjs
@@ -26,6 +26,16 @@ import { PathViolation, safeJoin } from "./workspace.mjs";
  */
 export const EVIDENCE_INLINE_LIMIT_BYTES = 256 * 1024;
 
+/** Hard ceiling for the repository-owned verification command (WM-262). */
+export const DEFAULT_REPO_VERIFY_TIMEOUT_MS = 120_000;
+
+function repoVerifyTimeoutMs() {
+  const configured = Number(process.env.FACTORY_REPO_VERIFY_TIMEOUT_MS);
+  return Number.isFinite(configured) && configured > 0
+    ? configured
+    : DEFAULT_REPO_VERIFY_TIMEOUT_MS;
+}
+
 export const REFUSAL_REASONS = [
   "missing_input",
   "permission_denied",
@@ -84,7 +94,17 @@ export { normalizeFailureOutput, failureSignature };
  *         | { kind: "completed", result: object, receipt: object }}
  * @throws {ContractViolation} on any contract failure — fail closed.
  */
-export function verifyResult({ spec, def, registry, workspaceDir, attempt, journalHead = null, extraArtifacts = [], worktreeRecord = null }) {
+export function verifyResult({
+  spec,
+  def,
+  registry,
+  workspaceDir,
+  attempt,
+  journalHead = null,
+  extraArtifacts = [],
+  worktreeRecord = null,
+  verifyTimeoutMs = repoVerifyTimeoutMs(),
+}) {
   let raw;
   try {
     raw = readFileSync(path.join(workspaceDir, "result.json"), "utf8");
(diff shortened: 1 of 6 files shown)


hdkiller:
Linear handoff is pending: the implementation and CI are green, but Linear is currently returning its organization-wide 2,500 requests/hour rate limit, so the required WM-262 handoff comment and In Review transition could not be posted.

Exact verification: `bun test event-runtime/lib/verify.test.mjs event-runtime/lib/workspace.test.mjs event-runtime/lib/worker.test.mjs` — 106 pass, 0 fail.

PR #456, opened 2026-08-17
feat(schedules): add factory supply loops

hdkiller:
## Summary
- add enabled, auto-approved `work-factory` and `triage-factory` supply schedules
- preserve disabled state for every non-factory repo schedule
- cover slot idempotency, catch-up suppression, and singleton planning for the shipped work loop

## Verification
- `bun test event-runtime/lib/schedules.test.mjs && bun event-runtime/cli.mjs schedule` — 27 tests pass; isolated runtime lists both loops enabled with correct next-due times

UX critique: skipped — backend scheduler configuration has no user-facing flow

Fixes WM-496

### DIFF
--- a/event-runtime/lib/schedules.test.mjs
+++ b/event-runtime/lib/schedules.test.mjs
@@ -471,6 +471,66 @@ describe("scheduleView (§9)", () => {
 });
 
 describe("registry validation of schedules.json", () => {
+  test("the factory supply loops are enabled, automatic, singleton schedules", () => {
+    expect(base.schedules["work-factory"]).toEqual({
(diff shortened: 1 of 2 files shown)


hdkiller:
CI exposed a deterministic Owned Paths closure gap: `event-runtime/loop.test.mjs` hard-codes that these schedules must not exist, but WM-496 permits changes only to `event-runtime/schedules.json` and `event-runtime/lib/schedules.test.mjs`. Closing rather than widening scope. Filed WM-497 with the exact failing assertions; this branch remains available once scope is resolved.

PR #495, opened 2026-08-17
chore(dispatch): raise dispatch run cap 45→90 min (WM-567)

hdkiller:
Fixes WM-567

Raises the dispatch agent's `limits.timeout_seconds` 2700 → 5400 and the policy backstop `limits.max_run_minutes` 45 → 90 (with comment), plus the matching comment in `event-runtime/lib/verify.mjs`.

Why: WM-470 run_5aec6675 (large architecture ticket) hit the 45-min cap right after pushing its PR and lost the run. Operator asked 2026-08-17 for longer timeouts for such jobs; a runtime extend feature is tracked separately (WM-566).

Verification: `bun event-runtime/cli.mjs update-pins` (pins already current) && `bun test event-runtime/lib/planner.test.mjs orchestrator/repos-config.test.mjs` — 39 pass, 0 fail.

## Handoff
UX critique: n/a (config only).

### DIFF
--- a/config/policy.yaml
+++ b/config/policy.yaml
@@ -130,7 +130,13 @@ limits:
   # Without it, raising a harness timeout to let long work finish also raises
   # how long a stuck run holds a concurrency slot — that trade is only
   # acceptable when something else guarantees the slot eventually frees.
-  max_run_minutes: 45
+  #
+  # 90 min: raised from 45 (WM-567) after WM-470 run_5aec6675 — a large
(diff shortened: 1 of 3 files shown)


hdkiller:
CI: Verify job has failed 4x (13:48–14:02 UTC) with a 429 from codeload.github.com while downloading actions/checkout on watt-mind-runner-6 — no repo code ran. Runner-infra issue tracked as WM-570; PR content is verified locally (39/39 tests, pins current). Needs a rerun once the rate limit clears.

PR #501, opened 2026-08-17
feat(event-runtime-web): linkify run trace refs (WM-546)

hdkiller:
## Summary
- linkify HTTP(S), GitHub PR, and Linear issue references in run-trace prose while preserving code/tool-input rendering
- format unknown or zero-with-cost usage as `tokens n/a` with two-decimal cost, including partial aggregate usage
- keep full-run trace and detail panes in normal stacked flow below the desktop breakpoint

## Verification
- `cd event-runtime/web && bun run build && bun test` — pass (680 tests)

UX critique: required — FIX-FIRST unresolved after 2 rounds. All WM-546 link and usage criteria passed in the running app at desktop and 1200px; the remaining 390px limitation is the persistent app-level sidebar outside Owned Paths and is tracked by WM-572. Evidence: http://127.0.0.1:7405/#/run/run_5fb19dca-6337-4e29-a7ff-dc75081aad35 and `wm546-390-mobile-links.png` in the run workspace.

Fixes WM-546

### DIFF
--- a/event-runtime/web/src/components/RunTrace.test.tsx
+++ b/event-runtime/web/src/components/RunTrace.test.tsx
@@ -3,7 +3,12 @@ import { afterEach, describe, expect, test } from "bun:test";
 import { act, cleanup, fireEvent, waitFor } from "@testing-library/react";
 import type { RunState, TraceEntry } from "../types";
 import { changeInput, renderWithClient, restoreApi } from "../test-render";
-import { formatErrorRowLabel, LIVE_STATES, RunTrace } from "./RunTrace";
+import {
+  formatErrorRowLabel,
+  formatUsageSummary,
+  LIVE_STATES,
+  RunTrace,
+} from "./RunTrace";
 
 afterEach(() => {
   cleanup();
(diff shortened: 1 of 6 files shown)


hdkiller:
## UX critique: FIX-FIRST

Rebased cleanly onto current develop (963 pass, 0 fail; build clean), then critiqued live.

**Finding (major, in-scope) — Linkified references in trace prose have no visual affordance distinguishing them from plain text.** In `RunTrace.tsx`, `renderLinkifiedText` styles every auto-linkified match (URL, GitHub PR ref, Linear issue ref) as `text-(--text-dim)`, identical to the surrounding paragraph text — confirmed via computed styles in both themes (dark: link `oklch(0.7144 0.00484 260)` == parent; light: link `oklch(0.427 0.0092 260)` == parent). No underline at rest either (only `hover:underline`). Reading trace prose containing refs like "WM-726/WM-730", it's impossible to tell those tokens are clickable without hovering over each word individually. This is inconsistent with the app's own established link styling elsewhere in the same file: markdown `[text](url)` links use `text-(--accent) underline` (always visible), and the existing table-cell PR/ticket links also render in accent blue. Suggested fix: give the linkify `<a>` a distinct at-rest color (e.g. `text-(--accent)` at reduced opacity, or reuse the markdown-link style) so refs are discoverable by scanning, not just hovering.

What worked well: linkification correctness is solid (GitHub PR refs and Linear issue refs both produce correct hrefs, verified across dozens of live occurrences); code/tool-input isolation works exactly as specified (code/pre blocks never get linkified, even when they contain ref-like text); usage/cost formatting is clean and correct ("tokens n/a · $0.00" for unknown/zero-cost, real numbers otherwise); responsive stacking is correct at 1200px/900px (detail pane drops below trace, not squeezed to a side column).

Branch rebased onto current develop and pushed (commit 6ba44eb) but left in draft — needs the link-affordance styling fixed before this can ship.

PR #505, opened 2026-08-17
fix(web): replace native Display option controls

hdkiller:
Fixes WM-556

Replaces the Display panel's native grouping and ordering selects with keyboard-navigable styled listboxes, adds a two-state ordering direction segment, and clarifies property chip states with checkmarks and pressed semantics.

UX critique: required — SHIP
Evidence: http://127.0.0.1:7747/#/runs; /var/folders/pd/h5j5j7953dx5h5l9q2jxdwch0000gn/T/pi-chrome-devtools-screenshot-0fd2d15b-8ca3-4e0a-b774-17b85cbad2fe.png

Verification: `cd event-runtime/web && bun run build && bun test` — 676 pass, 0 fail.

### DIFF
--- a/event-runtime/web/src/components/DisplayOptions.test.tsx
+++ b/event-runtime/web/src/components/DisplayOptions.test.tsx
@@ -71,8 +71,18 @@ afterEach(() => {
   goPrefix.armedAt = 0;
 });
 
+function chooseOption(
+  r: ReturnType<typeof render>,
+  label: string,
+  option: string,
+) {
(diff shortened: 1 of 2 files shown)


hdkiller:
CI environment note: the Verify job failed before checkout because GitHub codeload returned HTTP 429 while downloading `actions/checkout@v7` (3 attempts). Project code did not run; the exact local verification passed with 676 tests. Recent unrelated runs show the same ENV failure, so this should be rerun after the codeload throttle clears.

PR #516, opened 2026-08-17
fix(web): polish chain trace view

hdkiller:
Fixes WM-587

- fit the chain on initial layout, node growth, route changes, and reset
- prioritize distinguishing event labels and ordinal identical siblings
- compact the chain header and align detail-pane grammar
- strengthen selected-node highlighting
- add focused chain label and fit tests

UX critique: required — FIX-FIRST unresolved after two review rounds; round-2 viewport findings were addressed after the final allowed round, so this remains draft pending reviewer confirmation.

Verification: `cd event-runtime/web && bun run build && bun test` — pass (679 tests).

### DIFF
--- a/event-runtime/web/src/views/Chain.test.tsx
+++ b/event-runtime/web/src/views/Chain.test.tsx
@@ -1,14 +1,109 @@
 import "../test-dom";
-import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { useEffect } from "react";
+import {
+  afterAll,
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  mock,
+  test,
+} from "bun:test";
 import { cleanup, fireEvent, waitFor } from "@testing-library/react";
-import { Chain } from "./Chain";
 import { CHAIN_VIEW_STORAGE_KEY } from "../chainTimeline";
 import {
   createRunDetailFixture,
   renderWithClient,
   restoreApi,
+  withApi,
 } from "../test-render";
-import type { ChainView, LifecycleEvent, Proposal } from "../types";
+import type { ChainEvent, ChainView, LifecycleEvent, Proposal } from "../types";
+
+const fitView = mock((_options: unknown) => {});
+
+// `mock.module` persists for the whole `bun test` process, so this overrides
+// only the one export Chain's own tests actually need faked — the real
+// `<ReactFlow>` needs full viewport/DOM measurement this environment can't
+// give it. Everything else (Handle, Background, Position, ReactFlowProvider,
+// applyNodeChanges, …) stays the genuine module, so files that happen to run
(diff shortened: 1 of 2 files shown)


hdkiller:
## UX critique: FIX-FIRST

Rebased cleanly onto current develop (a large unrelated Timeline-mode feature, WM-639, was merged onto Chain.tsx/Chain.test.tsx since this branch diverged — both features now coexist; 958 pass, 0 fail, build clean), then critiqued live against a real 25-event/22-run chain with 16 duplicate sibling events (good stress case for the ordinal-labeling feature).

**Finding 1 (blocker, in-scope) — DetailPane breadcrumb title collapses to an identical string for sibling nodes it's supposed to distinguish.** Opening any of the 16 duplicate `merge-fix.requested` event nodes shows the DetailPane breadcrumb truncated to the shared run prefix (`Chain / planned chain-run_a2ced6fa`) -- verified identical `innerText` for two different nodes with different downstream proposals/runs. The graph canvas itself correctly labels these `merge-scan@2 #1` / `#2`, but that ordinal never makes it into the breadcrumb, which is exactly where a user checks "which node am I looking at."

**Finding 2 (major, in-scope, explicitly requested) — `?view=timeline&node=<id>` deep link does not work.** The code's own comment promises this query-spelling lifts the node into selection so the row highlights and the pane opens. Tested three times with different node ids: no highlight, no pane open, URL never normalizes. The equivalent path-segment form (`#/chain/<id>/<nodeId>?view=timeline`) works perfectly. Root cause looks like `onSelectNode`'s `navigate(hashPath(...))` building a fresh path and dropping the existing `?view=timeline` query, or the `?node=` mount effect not firing at all. Lower real-world blast radius since nothing in the app's own "Copy link" generates `?node=` links -- but the task explicitly asked to verify this exact deep-link and the source comment claims it works.

Minor/follow-up (non-blocking): sibling RUN nodes don't inherit the ordinal suffix their sibling EVENT nodes get (incomplete extension of the labeling work); node text is illegible at initial fit-to-viewport zoom on dense chains (expected trade-off, zoom controls work).

What worked well: initial graph fit handles even this large lopsided chain cleanly; ordinal labeling on EVENT nodes is correct (verified via DOM dump, #1-#16 no gaps/dupes); compact single-line header reads clearly in both themes; selected-node highlighting and "Show on canvas"/"Show in timeline" work; Graph/Timeline toggle (both `t` key and click) preserves selection; Timeline mode's own sibling disambiguation (PR/ticket context) is actually more informative than Graph's ordinal-only labels -- worth reusing for the breadcrumb fix; j/k narrative nav works; light theme clean; no console errors.

Branch rebased onto current develop and pushed (commit 9778cc1) but left in draft -- needs the breadcrumb and deep-link fixes before this can ship.

PR #558, opened 2026-08-17
feat(event-runtime-web): inbox-first Overview

hdkiller:
## Summary
- Lead Overview with a lazy, Inbox-owned Needs you section grouped by attention type.
- Fold runtime anomalies into the same compact row grammar and retain keyboard/open/ack affordances.
- Prefix the document title with the open Inbox count.

## Verification
- `cd event-runtime/web && bun run build && bun test` — pass (826 tests)
- `factory security` — pass

## UX critique
required — NOT-ASSESSED: Chromium tooling could not launch, so the running app could not be driven. This PR remains draft.

Fixes WM-596

### DIFF
--- a/event-runtime/web/src/App.tsx
+++ b/event-runtime/web/src/App.tsx
@@ -546,7 +546,9 @@ export function App() {
     const id = route.length > 1 ? route[route.length - 1] : null;
     const typeQ = hashSearch(window.location.hash).get("type");
     const detail = id ?? typeQ;
-    const pageTitle = detail ? `factory · ${viewLabel} · ${detail}` : `factory · ${viewLabel}`;
+    const pageTitle = detail
+      ? `factory · ${viewLabel} · ${detail}`
+      : `factory · ${viewLabel}`;
     document.title = inboxOpen > 0 ? `(${inboxOpen}) ${pageTitle}` : pageTitle;
   }, [inboxOpen, route, viewLabel]);
 
--- a/event-runtime/web/src/App.tsx
+++ b/event-runtime/web/src/App.tsx
@@ -546,10 +546,9 @@ export function App() {
     const id = route.length > 1 ? route[route.length - 1] : null;
     const typeQ = hashSearch(window.location.hash).get("type");
     const detail = id ?? typeQ;
-    document.title = detail
-      ? `factory · ${viewLabel} · ${detail}`
-      : `factory · ${viewLabel}`;
-  }, [route, viewLabel]);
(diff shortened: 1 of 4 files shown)


hdkiller:
## UX critique: FIX-FIRST

Rebased cleanly onto current develop (959 pass, 0 fail; build/tsc/eslint clean), then critiqued live against real data (18 open inbox items, 4 CI-red escalations, 2 runtime config anomalies).

**Finding 1 (major, in-scope) — Runtime anomalies render twice on the same screen in two different visual styles, one screenful apart.** The two config anomalies ("FACTORY_EVENT_SECRET is unset…", "FACTORY_GITHUB_WEBHOOK_SECRET is unset…") appear verbatim inside "Needs You → Runtime" (plain text, amber "Runtime" tag, no Open/ack) *and* again in the separate "Anomalies · 2 active issues" panel directly below (bordered box, "configuration" tag, "copy" button). Reads as a bug/duplicate-log to a first-time viewer. The ticket's own framing ("folds runtime anomalies into the same compact row grammar") suggests replacing the old Anomalies block, not appending to it — pick one: either Needs You's Runtime rows become the anomaly surface (drop or collapse the separate panel), or the Anomalies panel stays canonical and Needs You just badges a count/link into it.

**Finding 2 (major, in-scope) — The "18 open" header total silently excludes the Runtime subgroup with no visible explanation.** Decide (14) + Red (4) = 18 matches the header, the Inbox nav badge, and the Inbox view's own "Open 18" filter (all cross-checked and consistent). But the on-screen subgroups visibly read 14 + 4 + 2 = 20 against the "18 open" total shown directly above — nothing tells the reader Runtime doesn't count toward "open." Since every other number on the page sums correctly, this invites a "the count is wrong" read. Either fold Runtime into the total (if it should count) or visually separate/relabel so the header doesn't imply a sum of the groups below it.

Minor/follow-up (non-blocking): keyboard roving selection isn't native DOM focus (pre-existing app pattern, a11y follow-up); narrow-viewport (430px) loses row title text entirely rather than truncating (pre-existing shell assumption, no mobile requirement in this ticket); "9 more →" is a full navigation not an inline expand (acceptable, just a beat of hesitation).

What worked well: count cross-check is solid and live-accurate; keyboard nav (j/k, Enter) correctly roves only actionable rows and opens the real Inbox detail view; grouping by attention type (Decide/Red/Runtime) is clear and renders correctly in both themes.

Branch rebased onto current develop and pushed (commit 371cfac) but left in draft — needs the two major findings addressed before this can ship.

PR #569, opened 2026-08-18
fix(event-runtime-web): make list toolbars responsive (WM-543)

hdkiller:
## Summary
- add a shared responsive list-toolbar layout with non-wrapping tabs and a right-aligned tools group
- migrate Events, Proposals, Runs, Workers, Artifacts, Projects, Agents, and Schedules
- constrain filters to shrink from 14rem to 9rem before the tools group wraps
- add structural render coverage for one tablist and one tools group

## Verification
- `cd event-runtime/web && bun run build && bun test` — pass (843 tests, 0 failures)

UX critique: required — NOT-ASSESSED; the isolated Chromium browser exited before DevTools became available on both critique attempts, so the requested 1440/1100 screenshots could not be captured. PR intentionally remains draft.

Fixes WM-543

### DIFF
--- a/event-runtime/web/src/components/ui.test.tsx
+++ b/event-runtime/web/src/components/ui.test.tsx
@@ -348,7 +348,10 @@ describe("ListToolbar (WM-543)", () => {
     expect(classes(tools)).toContain("flex-nowrap");
     expect(classes(tools)).toContain("flex-[1_1_14rem]");
     expect(classes(tools)).not.toContain("w-max");
-    expect(classes(tools)).toContain("justify-end");
+    // Right-aligned only from `sm` up; below that, overflow runs off the
+    // trailing edge rather than back over the nav sidebar (WM-96 follow-up).
+    expect(classes(tools)).toContain("justify-start");
+    expect(classes(tools)).toContain("sm:justify-end");
     expect(classes(tools.parentElement as HTMLElement)).toContain("flex-wrap");
 
     const filter = r.getByRole("combobox").parentElement?.parentElement;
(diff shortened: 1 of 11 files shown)


hdkiller:
Fixed the review's blocking findings on `ListToolbar` (`event-runtime/web/src/components/ui.tsx`):

**1. Tab strip clipping (BLOCKING, WM-96 regression risk)** — at ~800-880px the tab strip scrolled but gave no visual affordance. Added a ResizeObserver-driven CSS `mask-image` edge fade (theme-aware, light/dark) that shows/hides per scroll position, plus `scrollIntoView` on tab focus for keyboard reachability. Kept the JSDoc guardrail comment on `ListToolbar` describing both the WM-543 (no-wrap) and WM-96 (no silent clip) constraints.

**2. Workers mobile `<select>` width (mechanical)** — restoring `w-full` alone didn't work: the tabs-wrapper ancestor is `shrink-0` with no explicit width, so it only ever hugs content and a percentage-width child has nothing to resolve against (same latent bug existed in Projects.tsx). Root-caused it in the shared component instead: the wrapper is now `w-full` below `sm` and `sm:w-auto sm:shrink-0` at/above it — fixes both Workers and Projects.

**3. Tools-row/sidebar overlap (found during critique, not in original review)** — at very narrow widths (~375px) the "List tools" row (Display/filter) overflowed past its own left edge onto the nav sidebar, because `justify-end` + `flex-nowrap` anchors overflow toward the start. First attempt added `overflow-x-auto` to contain it, but that had to be reverted: it forces `overflow-y` to compute as `auto` too (CSS overflow pairing rule), which clipped `FilterInput`'s plain `position:absolute` suggestion dropdown to a sliver. Fixed instead with `justify-start` below `sm`, `sm:justify-end` at/above — overflow now runs off the trailing edge instead of back over the sidebar, with zero change to desktop alignment and zero risk to the dropdown.

**Verification**
- `bun run build` + `bun test` — 950 pass, 0 fail
- `bunx prettier --write` + `bunx eslint --fix` on changed files; `bun run lint` clean (only pre-existing unrelated warnings)
- Two rounds of `factory-ux-critic` browser verification against the live dev server:
  - Round 1 caught that the Workers select fix didn't actually take effect and found the sidebar-overlap regression — FIX-FIRST.
  - Round 2 verified all three fixes (tab-strip fade across 800/880/1280px × light/dark with focus-scroll, Workers select at 375/600px, tools-row overlap gone, filter-dropdown NOT clipped at 375px or 1280px, desktop right-alignment unchanged) — **verdict: SHIP**.

Marking ready for review.

PR #573, opened 2026-08-18
fix(event-runtime-web): preserve concurrent requeue polls

hdkiller:
## Summary
- remove the component-wide alive flag that cancelled every requeue confirmation after the first navigation
- let each poll independently reach proposal match, timeout, or network-error feedback
- add hook and Overview bulk-requeue regression coverage for concurrent polling across unmount

## Verification
- `cd event-runtime/web && bun test src/hooks.test.ts src/hooks.test.tsx src/views/Events.test.tsx src/views/Overview.test.tsx` — 72 pass, 0 fail
- `cd event-runtime/web && bun run build` — pass

UX critique: required — NOT-ASSESSED; worktree and app health passed, but Chromium exited before DevTools became available, so the critic could not drive the flow or collect browser evidence. PR remains draft.

Fixes WM-247

### DIFF
--- a/event-runtime/web/src/hooks.test.ts
+++ b/event-runtime/web/src/hooks.test.ts
@@ -6,7 +6,7 @@ import {
   useQuery,
 } from "@tanstack/react-query";
 import { act, cleanup, render, waitFor } from "@testing-library/react";
-import { createElement as h, useState } from "react";
+import { createElement as h, useEffect, useState } from "react";
 import { api } from "./api";
 import {
   modal,
@@ -169,8 +169,13 @@ function ThemeProbe() {
 type PollRequeue = ReturnType<typeof useRequeuePoll>;
 let exposedPollRequeue: PollRequeue | null = null;
 
-function RequeuePollProbe(props: { onJumpProposal: (proposalId: string) => void }) {
-  exposedPollRequeue = useRequeuePoll(props.onJumpProposal);
+function RequeuePollProbe(props: {
+  onJumpProposal: (proposalId: string) => void;
+}) {
+  const pollRequeue = useRequeuePoll(props.onJumpProposal);
+  useEffect(() => {
+    exposedPollRequeue = pollRequeue;
+  }, [pollRequeue]);
   return null;
 }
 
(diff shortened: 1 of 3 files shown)


hdkiller:
Code review verdict: **FIX** (factory-merge-reviewer)

CI `Verify` is red — the diff introduces 3 eslint errors in its new test scaffolding (production change itself is correct; rebase is clean, `mergeable: MERGEABLE`, sits directly on current develop tip).

Reproduced locally (`bun run lint`): 3 errors, everything after Lint (typecheck/build, tests, E2E) was skipped by CI as a result.

1. `event-runtime/web/src/hooks.test.ts:173:3` — `react-hooks/globals`: `exposedPollRequeue = useRequeuePoll(props.onJumpProposal);` reassigns a module-scope `let` during render (a render side effect). Fix: pass a ref/holder as a prop and assign it in a `useEffect` instead.
2. `event-runtime/web/src/hooks.test.ts:200:11` — `prefer-const` on `let view!: ReturnType<typeof render>;` (assigned exactly once). Drop the predeclaration, declare at the assignment site.
3. `event-runtime/web/src/views/Overview.test.tsx:726:11` — same `prefer-const` pattern with `renderOverview`.

After fixing, re-run `bun run lint` — the ticket's verification command doesn't cover eslint, which is why this reached CI red instead of being caught locally.

Non-blocking notes for follow-up (not required for this PR):
- Bulk-requeue can now chain-navigate through up to N proposal routes across the 8s window since the shared `alive` flag is gone (this is the intended WM-247 behavior, not a defect, but worth a UX follow-up: jump on first match, toast subsequent matches).
- The Overview test's shared `api.proposals` mock could pick up the deck's own live poll under load — a minor flake vector (`expect(pollRequests.length).toBeGreaterThanOrEqual(2)` would be safer than an exact-length wait).
- Requeue toast still embeds the full raw eventId (pre-existing `notify()` behavior, already flagged by the UX critic).
- PR body/draft state should be refreshed — it still says "UX critique: required — NOT-ASSESSED" though a SHIP verdict is now posted on WM-247.

Leaving this in draft per process; not fixing in this pass.

PR #575, opened 2026-08-18
feat(event-runtime-web): unify control sizes and tooltips (WM-589)

hdkiller:
## Summary
- add shared 24/28/32px Button sizing and a mandatory-label, tooltip-backed IconButton
- route production controls through the shared primitive and migrate Runs pane/trace/theme/repo icon actions
- make Runs tab counts follow active filters and align the empty-state Clear filter action
- add regression coverage for control sizing, IconButton accessibility, shared tooltips, and filtered counts

## Verification
- `cd event-runtime/web && bun run build && bun test` — pass (846 tests, 0 failures)

UX critique: required — NOT-ASSESSED; two attempts could not launch the isolated Chromium DevTools browser, so no visual approval or screenshot evidence was available. PR intentionally remains draft.

Fixes WM-589

### DIFF
--- a/event-runtime/web/src/App.tsx
+++ b/event-runtime/web/src/App.tsx
@@ -41,12 +41,13 @@ import {
   useGoSequences,
   type PaletteAction,
 } from "./components/CommandPalette";
-import { ToastContainer, copyLink, copyText } from "./components/ui";
+import { IconButton, ToastContainer, copyLink, copyText } from "./components/ui";
 import type { ArtifactFilters } from "./views/Artifacts";
 import { Events } from "./views/Events";
 import { Overview } from "./views/Overview";
 import { Runs } from "./views/Runs";
 import { NAV, type NavKey } from "./nav";
+import { Button as PrimitiveButton } from "./components/ui";
 
 type WorkerHealthFilter = "live" | "busy" | "stale";
 const isWorkerHealthFilter = (
@@ -619,7 +620,8 @@ export function App() {
               />
               <span className="display text-[14px] font-semibold">factory</span>
             </div>
-            <button
+            <PrimitiveButton
+              bare
               type="button"
               className="rounded px-1.5 py-0.5 text-[11px] font-semibold tracking-wide uppercase"
               title={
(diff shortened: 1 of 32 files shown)


hdkiller:
UX critique verdict: **FIX-FIRST** (factory-ux-critic)

**#1 — blocker (in-scope): the theme-toggle tooltip renders completely off-screen, on every window size.**
The footer theme toggle (`App.tsx`, now an `IconButton`) lives in a status bar pinned to the bottom of the viewport. The new shared `Tooltip` (`ui.tsx`) always opens downward (`top-full`) with no viewport-collision handling. Verified via focus + `getBoundingClientRect()`: the tooltip box sits at `top: 905.5` in a 900px-tall viewport — it computes `opacity: 1` but renders entirely below the visible page, on any screen height, since the footer is pinned to the bottom edge. The previous native `title` attribute auto-flipped to stay on-screen; the new custom tooltip has no fallback — a straight regression on the exact control the PR calls out as primary to verify.

**#2 — major (in-scope): same `Tooltip` has no horizontal edge-awareness — clips off the right side for right-aligned icon buttons.**
The ContextTabs "+" repo-picker trigger sits top-right. Its tooltip centers under the trigger (`left-1/2 -translate-x-1/2`) with no flip/clamp logic — on a 1440px viewport it extends to `right: 1487.7`, ~48px past the edge, visibly truncating "Open a repo tab (g 1–9)". Same root cause as #1: the shared `Tooltip` primitive needs basic viewport-collision handling (clamp/flip) before it becomes the default for every icon-only control in the app — any future edge-placed IconButton will hit this.

**#3 — minor follow-up:** the repo "+" trigger loses its visible "Repo" text label (intentional per PR body); combined with #2, a first-time user currently has no way to learn what "+" does until #2 is fixed. Not a blocker on its own.

**What worked well:** Runs tab counts responding to an active filter read clearly (verified with a real filter query). IconButton+Tooltip render correctly everywhere away from screen edges (Run detail copy-id/CLI/link, dark+light). No sign of the rebase risk flagged going in — toured Overview, Runs (list+detail), Workers (list+detail), ContextTabs in both themes with zero dropped/misplaced/duplicated UI elements.

Screenshots under the worktree's `.factory/critique/` (11 files).

Leaving in draft per process; not fixing in this pass.

PR #579, opened 2026-08-18
fix(event-runtime): expose RunSpec on runs list rows

hdkiller:
## Summary
- include the full parsed RunSpec on every `GET /runs` row while preserving flattened fields
- type run list rows with the existing `RunSpec` and update representative fixtures
- cover API row shape plus Runs custom-column rendering and field discovery

## Verification
- `bun test event-runtime/lib/api-runs.test.mjs event-runtime/lib/api.test.mjs event-runtime/web && cd event-runtime/web && bun run build`

UX critique: required — NOT-ASSESSED (isolated Chromium exited before DevTools became available; PR remains draft)

Fixes WM-303


hdkiller:
Code review verdict: **FIX** (factory-merge-reviewer)

Diff is correct and well-scoped, rebase preserved develop's concurrent work intact (WM-725 test additions in Runs.test.tsx/Workers.test.tsx fully intact, no dropped assertions), but the new WM-303 test block in `event-runtime/web/src/views/Runs.test.tsx` (lines ~698, 706, 708, 709) is not Prettier-formatted, which will fail CI's required `Verify` job at the formatting-check step. `bun run format` fixes it — no other file in the diff has a violation.

Non-blocking, worth a Triage follow-up: `GET /runs` payload grows ~3-4x on an unpaginated, 2s-polled list now that the full RunSpec is included — measured against the live runtime DB (289 runs): spec sizes avg 1138B, max 11223B, 329KB total, not "a few hundred bytes" as the ticket assumed, and the list has no pagination/LIMIT. Not blocking on localhost but worth a follow-up (LIMIT/`?fields=` on `GET /runs`, or omitting spec for terminal rows).

Also non-blocking: schema discovery only walks `spec.input.*`, not all of `spec.*` as the ticket's AC implies (e.g. `spec.timeoutSeconds`, `spec.placement.*` render fine as custom columns but never appear in the discovery dropdown) — scope note, not a defect in what's implemented and tested.

PR is still draft with UX critique NOT-ASSESSED from before a Chromium fix landed on develop — the orchestrator's backend/test-only classification (no `.tsx` source changed, only test-render.tsx and types.ts) is defensible, but note the Display->"Add custom property path" dropdown now surfaces new `spec.input.*` entries which is an operator-visible effect delivered purely through data.

No security/escalation concern — spec was already served in full by `GET /runs/:id` and `GET /proposals`; no new data class exposed (verified no credentials/tokens in the 289 live spec payloads).

Leaving in draft per process; not fixing in this pass.

PR #580, opened 2026-08-18
fix(event-runtime-web): align detail pane headers

hdkiller:
## Summary
- extract shared `PaneHeader` chrome for consistent identity, verb, and copy rows
- align Agents, Schedules, and Workers detail panes to 440px breadcrumb headers
- compact Workers adapters and labels and humanize process field names
- add breadcrumb, width, and worker-detail regression coverage

## Verification
- `cd event-runtime/web && bun run build && bun test` — pass (845 tests)

UX critique: required — NOT-ASSESSED (isolated Chromium exited before DevTools became available; PR remains draft)

Out-of-scope canonical Proposals/Runs 460px width drift filed as WM-685.

Fixes WM-552

### DIFF
--- a/event-runtime/web/src/components/ui.tsx
+++ b/event-runtime/web/src/components/ui.tsx
@@ -687,12 +687,16 @@ export function PaneHeader({
     <div className="shrink-0 border-b border-(--border) px-4 py-3">
       {/* Row 1: Title & Close */}
       <div className="flex items-center justify-between gap-2">
-        <div className="display min-w-0 flex-1 truncate text-[14px] font-semibold">{title}</div>
+        <div className="display min-w-0 flex-1 truncate text-[14px] font-semibold">
+          {title}
+        </div>
         {close != null && <div className="shrink-0">{close}</div>}
       </div>
       {/* Row 2: Verb Row (≤ 3 bordered buttons) */}
       {actions != null && (
-        <div className="mt-2 flex flex-wrap items-center justify-between gap-1.5">{actions}</div>
+        <div className="mt-2 flex flex-wrap items-center justify-between gap-1.5">
+          {actions}
+        </div>
       )}
       {/* Row 3: Utility Row (copy/share quiet text line) */}
       {utility != null && (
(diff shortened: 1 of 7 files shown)


hdkiller:
UX critique verdict: **FIX-FIRST** (factory-ux-critic)

**Blocker (in-scope):** the Workers detail-pane breadcrumb title renders an identical, non-unique id for every worker spawned by the same supervisor process — `shortId()` (event-runtime/web/src/components/ui.tsx) assumes a single-underscore `prefix_hexbody` format (correct for run ids like `run_b0b1bf9e…`), but worker ids are `worker_<pid>_<hex>` (two underscores). Splitting on the *first* underscore and keeping 8 chars of the remainder lands exactly on the pid segment for every worker of this shape.

Verified live: `worker_2909182_2e08ba86` -> breadcrumb shows `worker_2909182_`, and a different worker `worker_2504903_474a477f` (different pid) -> breadcrumb shows `worker_2504903_`. Every worker under one supervisor PID gets the exact same visible breadcrumb text, with a trailing underscore that reads as cut off. This is new code from this PR (Workers.tsx, `shortId(sel.workerId)` in the new breadcrumb) — the list view still shows the full id correctly and only the new breadcrumb call is broken. Practical impact: you cannot tell which worker you're looking at from the breadcrumb (or a copied link's title) when the fleet shares a supervisor.

Fix: don't reuse `shortId()` for worker ids as-is — either fix it to split on the *last* underscore for this id shape, or use a different truncation for worker ids. The full id is already correctly available via the `title` tooltip and the Process section KV row, so only the breadcrumb's visible text needs fixing.

Minor follow-up (non-blocking): `shortId()`'s doc comment only documents the `type_hexbody` single-underscore case; worth a one-line note flagging it as unsafe for multi-underscore ids so the next caller doesn't hit the same trap.

Everything else checked out clean: Agents/Schedules breadcrumbs correct, all three panes measured exactly 440px, Workers adapter pills/humanized labels/empty placement-labels state all correct, no visual regressions dark/light, no console errors.

Leaving in draft per process — not fixing in this pass.

PR #585, opened 2026-08-18
feat(event-runtime-web): add operational metrics page

hdkiller:
## Summary
- add accessible, zero-dependency Sparkline, StackedBars, and p50/p95 Band SVG primitives
- add the live Metrics page with reliability, latency, spend, and approval sections plus persisted windows
- add truthful time/population drill-down contracts for Runs and Proposals and wire `#/metrics` / `g m`

## Verification
- `bun test event-runtime/web event-runtime/lib/api-runs.test.mjs event-runtime/lib/api-proposals.test.mjs` — 878 pass, 0 fail
- `cd event-runtime/web && bun run build` — pass

UX critique: required — NOT-ASSESSED (isolated Chromium exited before Chrome DevTools became available; no browser approval was inferred). PR intentionally remains draft.

Fixes WM-282

### DIFF
--- a/event-runtime/lib/api-proposals.test.mjs
+++ b/event-runtime/lib/api-proposals.test.mjs
@@ -1,0 +1,86 @@
+import { describe, expect, test } from "bun:test";
+import { makeServer } from "./api-test-helpers.mjs";
+
+describe("metrics proposal drill-down filters (WM-282)", () => {
+  const now = Date.parse("2026-08-18T10:00:00.000Z");
+  const from = "2026-08-18T08:00:00.000Z";
+  const to = "2026-08-18T10:00:00.000Z";
+
+  function insertProposal(
+    db,
+    id,
+    { status, createdAt, decidedAt = null, ttl = 3600 },
+  ) {
+    db.query(
+      `INSERT INTO proposals
+         (id, event_source, event_id, decision, status, created_at, ttl_seconds, decided_at, spec_json)
+       VALUES (?, 'test', ?, 'run', ?, ?, ?, ?, '{}')`,
+    ).run(id, `event-${id}`, status, createdAt, ttl, decidedAt);
+  }
+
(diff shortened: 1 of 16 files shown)


hdkiller:
## UX critique: FIX-FIRST

Rebased cleanly onto current develop (984 pass, 0 fail; build clean), then critiqued live.

**Blocker finding — 7d and 30d metrics windows return fabricated/misleading historical data.**
Verified at the API level: `GET /api/metrics?window=30d&bucket=24h` returns 30 daily buckets, but every series (`runs.outcomes`, `runs.started`, `spend.cost`, `spend.tokens`, `proposals.decisions`, `events.intake`, `attempts.retries`) is a flat zero for all 29 earlier buckets — the entire preceding-24h total is dumped into just the last bucket. The `7d` window (`bucket=6h`) shows the identical pattern. Visually this renders as a single collapsed spike bar at the right edge with a flat empty history. This is a backend aggregation bug for non-24h windows (the frontend correctly requests/renders whatever it's given — window/bucket params, axis labels, and localStorage persistence all work correctly). Given this page's whole point is historical reliability/latency/spend/approval trends, this needs a real fix before merge.

Minor/non-blocking follow-up: the Approval gate's "Time to decision" headline shows "0ms" while its own band chart directly below shows a real non-zero p50→p95 spike — plausible (median pulled down by auto/policy-approved decisions) but reads as broken at a glance; consider a clearer label.

What worked well: all four sections render real live data cleanly in both themes at 24h default; Sparkline/StackedBars/Band primitives render without glitches; drill-down from the Outcome-mix chart into `#/runs` correctly scopes via a "Metrics drill-down" banner; time-window buttons + `g m` shortcut + localStorage persistence all work correctly.

Branch was rebased onto current develop and pushed (commit 78ca86f) but left in draft — needs the 7d/30d aggregation bug fixed before this can ship.

PR #586, opened 2026-08-18
feat(event-runtime): require inbox resolve reasons

hdkiller:
## Summary
- add a retry-safe schema migration and ledger projection for persisted inbox resolution reasons
- require a non-empty reason in the resolve API and return it from GET /inbox
- add required single/bulk resolve reason controls, templates, and resolved-reason detail display

## Verification
- `bun test event-runtime/lib/inbox.test.mjs event-runtime/lib/api-inbox.test.mjs && cd event-runtime/web && bun test src/views/Inbox.test.tsx`
- `cd event-runtime/web && bun run build`

UX critique: required — NOT-ASSESSED; isolated Chromium exited before DevTools became available, so the live flow could not be visually driven. PR remains draft.

Fixes WM-604

### DIFF
--- a/event-runtime/lib/api-inbox.mjs
+++ b/event-runtime/lib/api-inbox.mjs
@@ -70,7 +70,14 @@ export async function handleInboxApiRoute({
     ) {
       return send(422, { error: "body must be an object" });
     }
-    if (
+    if (verb === "resolve") {
+      if (
+        typeof parsed.value.reason !== "string" ||
+        parsed.value.reason.trim() === ""
+      ) {
+        return send(422, { error: "reason must be a non-empty string" });
+      }
+    } else if (
       parsed.value.reason !== undefined &&
       typeof parsed.value.reason !== "string"
     ) {
@@ -80,7 +87,11 @@ export async function handleInboxApiRoute({
       const item =
         verb === "ack"
           ? ackInboxItem(db, id, { now: nowMs })
-          : resolveInboxItem(db, id, { now: nowMs, resolvedBy: actor });
+          : resolveInboxItem(db, id, {
+              now: nowMs,
+              resolvedBy: actor,
+              reason: parsed.value.reason.trim(),
+            });
       return send(200, { item });
     } catch (err) {
       const status = String(err.message).startsWith("unknown inbox item")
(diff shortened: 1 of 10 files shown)


hdkiller:
## Code review: FIX

Rebased cleanly onto current develop (migration correctly renumbered v7, no collision; 13 pass on inbox.test.mjs/api-inbox.test.mjs; 29 pass on Inbox.test.tsx; build clean). No UX critique was run initially (backend/schema-focused class of change) — but the review caught a real interaction bug in exactly the area the ticket's own Handoff flags as unassessed.

**Blocking finding — Enter key in the new resolve-reason textarea submits the dialog instead of inserting a newline.** The window-level `keydown` handler in `Inbox.tsx` treats plain `Enter` as "submit resolve" and doesn't check `isTypingTarget(e.target)` (a guard this same file already defines and uses for other shortcuts). The reason field is a multi-line `<textarea rows={3} resize-y>` — once an operator types any non-empty text and presses Enter to start a second line, the dialog submits immediately with the partial text instead of inserting a newline. This defeats the textarea's multi-line design and risks submitting an incomplete audit-record reason. The PR's own reference model (`Proposals.tsx`'s reject dialog, which the ticket says to follow) avoids this by using a single-line `<input>` with a scoped `onKeyDown`, not a global window listener. Fix: either match the reference pattern (single-line input, scoped Enter handler) or keep the textarea but guard the window handler with `isTypingTarget` and require Cmd/Ctrl+Enter to submit while focus is in it. Not caught by existing tests since they dispatch Enter on `document.body`, not the textarea.

**Also flagged**: the ticket's own Handoff comment states the UX critique is required and was NOT-ASSESSED ("review the single/bulk dialog keyboard and focus behavior first") — which is exactly the area with the bug above. Worth a real UX critique pass (or explicit sign-off) before merge, not just this code-level fix.

Migration/validation checks all passed: v1-v7 sequential with no gaps/duplicates, v7 is retry-safe (checks PRAGMA table_info before ALTER TABLE), CURRENT_SCHEMA_VERSION derives automatically, and develop's resolver-validation/decision-pending gates in `resolveInboxItem` are fully preserved.

PR remains draft pending the Enter-key fix.

PR #602, opened 2026-08-18
fix(web): SHA click opens artifact inspector (WM-699)

hdkiller:
## Problem

On `#/artifacts`, the SHA was the one obvious thing to click — and clicking it fired a browser file download instead of opening the split-pane inspector. The SHA text was wrapped in `<a href={artifactUrl(...)} download>` with `stopPropagation()`, which intercepted the click before the row's selection handler ever saw it.

## Change

- **The SHA is now a deep link into the inspector.** `<a href="#/artifacts/<sha256>">`, no `download`, no `stopPropagation()` — a left-click selects the row and opens the split pane, and because it is a real link, middle-click, Cmd-click and "Copy link address" all still do the right thing.
- **The row `onClick` skips modified clicks.** Without this, a Cmd-click that opens a background tab would also navigate the current one out from under the operator.
- **Downloading moved to an explicit per-row control** — a small `↓` link next to the SHA with `aria-label="Download <name>"` and `stopPropagation()`, so it downloads without selecting the row.
- **Rows are keyboard reachable**: `tabIndex={0}`, a `focus-visible` ring, and Enter/Space selects. The key handler is scoped to the row itself so Enter on an inner link or button still behaves normally.
- **Non-text artifacts no longer render as a page of U+FFFD.** When the fetched bytes contain NUL or decode with a high replacement-character ratio, the inspector shows file name, size and kinds plus a prominent "Download file" button, and drops the content-search box that would have nothing to search.
- The inspector toolbar keeps its primary `Download` action, `Copy Raw Content`, `Copy SHA-256`, and the `ArtifactPanel` Raw/View toggle — untouched. The toolbar download name is now derived from a shared `downloadName()` helper rather than being rebuilt inline.

## Verification

```
cd event-runtime/web && bun test src/views/Artifacts.test.tsx
```

14 pass, 0 fail, 80 `expect()` calls. Full web suite is also green (874 pass, 0 fail across 59 files), plus `bun run format:check`, `bun run lint` (0 errors) and `tsc --noEmit`.

**Negative testing:** all 7 new/changed assertions were observed failing against the pre-fix component (7 pass / 7 fail) before `Artifacts.tsx` was touched — the SHA link test failed because the link's accessible name was `Download artifact <sha>` and its `href` pointed at `/api/artifacts/...`, which is exactly the reported bug.

Five new regression tests cover: the SHA href and the absence of a `download` attribute plus the click opening the inspector; the row download control not selecting the row; Tab focus and Enter/Space selection; a modified click being left to the browser; and the binary fallback showing metadata and a download instead of a text preview.

## UX critique

Required, but **NOT-ASSESSED** — the `cursor-ide-browser` MCP bridge returned `No browser tab available` for every call (`browser_navigate` with and without `newTab`, `browser_tabs list`, `browser_cdp`), reproduced in both the critic subagent and this session, while the app itself was healthy (`http://127.0.0.1:7751/` → 200, seeded inventory served). Filed as WM-726. The interaction contract is covered by the happy-dom tests above; what went unreviewed is the purely visual side — the `↓` icon's clarity and hit target, and the focus-ring appearance.

## Risks

Look first at the `↓` download affordance: it is a 20×20 icon-only control, so if it reads as ambiguous or too small in the browser that is a follow-up polish item rather than something the tests can catch. Second, `looksBinary()` uses a heuristic (NUL byte, or >2% U+FFFD in the first 4KB) — a text artifact that legitimately contains replacement characters above that ratio would be shown as non-previewable, with the raw bytes still one download away.

Fixes WM-699

### DIFF
--- a/event-runtime/web/src/views/Artifacts.test.tsx
+++ b/event-runtime/web/src/views/Artifacts.test.tsx
@@ -163,7 +163,7 @@ describe("Artifacts inventory (WM-207)", () => {
     }
     expect(view.queryByRole("columnheader", { name: "Orphan" })).toBeNull();
     const download = view.getByRole("link", {
-      name: `Download artifact ${"a".repeat(64)}`,
+      name: "Download aaaaaaaaaaaa.report",
     });
     expect(download.getAttribute("href")).toContain(
       `/api/artifacts/${"a".repeat(64)}`,
@@ -179,8 +179,8 @@ describe("Artifacts inventory (WM-207)", () => {
     const view = renderArtifacts();
     await waitFor(() => expect(view.getByText("aaaaaaaaaaaa")).toBeTruthy());
 
-    const downloads = view.getAllByRole("link", { name: /Download artifact/ });
-    expect(downloads.map((link) => link.textContent)).toEqual([
+    const shas = view.getAllByRole("link", { name: /^[0-9a-f]{12}$/ });
+    expect(shas.map((link) => link.textContent)).toEqual([
       "cccccccccccc",
       "bbbbbbbbbbbb",
       "aaaaaaaaaaaa",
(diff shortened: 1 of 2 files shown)


hdkiller:
factory-merge-fix round=1 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=a618f515a3088c6567aa83e0b556346a50fef806 new=fe5638c72da3e8145d2c133d8a84ac10e0e9ba7e

PR #617, opened 2026-08-18
ui(event-runtime-web): show full repository config

hdkiller:
## Summary
- widen the allow-listed repository registry view with merge/escalation, owned-path, deploy/smoke, security, and effective concurrency settings
- group the Projects detail panel by operational concern with source annotations, explicit escalation states, and untruncated policy lists
- keep the responsive detail flow readable and update API/UI tests plus the web UI specification

## Verification
- `bun test event-runtime/lib/repos.test.mjs event-runtime/web/src/views/Projects.test.tsx && (cd event-runtime/web && bun run build)` — pass (43 tests; TypeScript and Vite production build green)
- `bun test event-runtime/lib/api-registry.test.mjs event-runtime/lib/api-config.test.mjs` — pass (13 tests)

UX critique: required — SHIP after 2 rounds; evidence: `http://127.0.0.1:7709/#/projects/factory`, `/tmp/WM-703-round2-narrow.png`, `/tmp/WM-703-round2-narrow-merge-gate.png`

Reviewer focus: `repos.mjs` imports the planner's exported default so the effective value cannot drift; WM-755 tracks moving that binding to a dependency-neutral module outside this ticket's Owned Paths.

Fixes WM-703

### DIFF
--- a/docs/event-runtime-webui.md
+++ b/docs/event-runtime-webui.md
@@ -1292,13 +1292,20 @@ worktrees a repo owns.
   team, and GitHub slug. `j`/`k` move, `c` copies the repo name,
   `#/projects/:name` is shareable per §10.8, and the empty state names the
   file that is empty (`config/repos.yaml`) rather than saying "no results".
-- **Detail panel.** **Configuration** — path (click to copy), GitHub link,
-  base and deploy branch, execution mode spelled out ("Autonomous
-  Dispatchable" / "Report Only (Watched)"), max in flight, worktree root, and
-  the verify command verbatim, since a paraphrased verify command is worse
-  than none. **Worktree Automation Scripts** renders `up`/`down`/`warm` as
-  three present-or-absent tiles: the shape of a repo's automation is
-  something you check before dispatching to it, not after.
+- **Detail panel.** Repository settings are grouped by what they govern, and
+  every group names `config/repos.yaml` as its source. **Dispatch** shows the
+  dispatchable/report-only mode, effective max in flight and whether it came
+  from the repo or the planner default, base branch, worktree root and script
+  capabilities, and the verify command verbatim. **Merge gate** shows the
+  `merge_ci` workflow and required checks plus every `escalate_paths` glob;
+  missing `escalate_paths` warns that every PR escalates, while an explicit
+  empty list says so. **Owned paths policy** shows direct source/required-path
+  rules and pin manifests, or the default. **Deploy & smoke** shows deploy
+  branch, the allow-listed deployment URL/branch/revision field, smoke
+  workflow/URL/deadline. **Security** currently allow-lists only
+  `python_version`, or states that defaults apply. Long path/check lists are
+  untruncated monospace chips inside their own scrolling containers, and all
+  other absent values use the shared empty-value placeholder.
 - **Janitor, Dry before Apply (OPS-362).** `POST /repos/:name/janitor` — §2's
   one mutation exception, specified in §7 — is wired as two buttons in an
   order the UI enforces rather than suggests: **Run Dry Janitor** first, and
--- a/event-runtime/lib/api-registry.test.mjs
+++ b/event-runtime/lib/api-registry.test.mjs
@@ -97,10 +97,23 @@ describe("agent and repository registry surfacing (OPS-212)", () => {
         deployBranch: "master",
         reportOnly: false,
         maxInFlight: 20,
+        effective: {
+          maxInFlight: 20,
+          maxInFlightSource: "repo",
+        },
         smokeDeadlineSeconds: null,
+        smokeWorkflow: null,
+        smokeUrl: null,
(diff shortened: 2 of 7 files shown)


hdkiller:
## Merge review: FIX

**Blocking — `effective.maxInFlight` reports the wrong number for repos with no `max_in_flight` in `config/repos.yaml`.**

`event-runtime/lib/repos.mjs` computes:
```js
effective: {
  maxInFlight: repo.maxInFlight ?? DEFAULT_MAX_IN_FLIGHT,
  maxInFlightSource: repo.maxInFlight === null ? "default" : "repo",
}
```
but the planner's real resolution (`planner.mjs:538`) is:
```js
const cap = repo.maxInFlight ?? maxInFlightFallback ?? policyMaxInFlight();
```
`policyMaxInFlight()` reads `config/policy.yaml → concurrency.max_in_flight_per_repo` (currently **50**, explicitly documented as "the FALLBACK; a repo's own max_in_flight in repos.yaml wins") and only falls through to `DEFAULT_MAX_IN_FLIGHT` (3) when that key is absent/invalid. The planner labels this source `"policy.yaml: max_in_flight_per_repo"`, never `planner.mjs`.

Repos with no `max_in_flight` set (coach-wattz, watts-mobile, hdkiller, eslint-config) will render `3 · default · planner.mjs` in this PR's UI while the planner actually admits up to 50 in-flight for them — i.e. the one field whose whole job is "the number the planner actually uses" is wrong for those repos, both the number and the source label.

Fix: resolve the fallback the same way the planner does (e.g. export `policyMaxInFlight` from `planner.mjs` and call it from `loadRepos({ root })`, storing the resolved value on the repo record for `reposView` to project — `reposView` itself takes no `root` so it can't read `policy.yaml` directly without breaking test fixture isolation). Widen `maxInFlightSource` to include a `"policy"` case, and update `repos.test.mjs` / `api-registry.test.mjs` fixtures to cover a repo that falls through to `concurrency.max_in_flight_per_repo`.

Minor / file-to-Triage (not blocking, for awareness — full detail in review transcript, happy to expand on request):
- `security` allow-list drops non-`python_version` keys (e.g. `hdkiller`'s `semgrep_args`), so the panel shows "defaults" for a repo that does have security config.
- New strict validation on `escalate_paths` in `loadRepos` means every `loadRepos` consumer (not just the escalation gate) now hard-fails on a malformed value — worth a conscious call, not obviously wrong.
- `repos.mjs` now imports from `planner.mjs`'s territory / vice versa creates an import-cycle risk already tracked as WM-755 — verified runtime-safe today, but fixing the maxInFlight finding above via an export deepens that edge.
- `api.ts`'s new `RepoItem` fields are optional even though the server always sends them; `undefined` would silently read as "escalate_paths not declared".
- `Projects.tsx` list `key={value}` could collide on duplicate globs/checks; a couple of hardcoded `"—"` dashes remain instead of the shared empty-placeholder.

CI: required `Verify` check has not reported yet (run still queued/pending at review time) — please confirm green before re-requesting review. Branch is mergeable/clean vs develop, Owned Paths clean, escalate_paths clears (factory repo has `escalate_paths: []`), no auth/payments/secrets/migration behavior touched.

hdkiller:
Fixed the blocking finding from review:

**`effective.maxInFlight` fallback** — `repos.mjs` now resolves the fallback through `planner.mjs`'s own logic instead of re-deriving one. Exported `policyMaxInFlightWithSource()` from `planner.mjs` (same resolution `worktreeDispatchAutoEligibility` uses); `loadRepos()` calls it once per load and carries the resolved value/source on each repo record, so `reposView()` projects the exact number the planner would actually admit — `repo` when `max_in_flight` is set, else the real `policy.yaml` value (`concurrency.max_in_flight_per_repo`, e.g. 50) with source `policy`, only falling further to `DEFAULT_MAX_IN_FLIGHT`/`default` when that key itself is absent/invalid. Added tests in `repos.test.mjs` and `api-registry.test.mjs` covering a repo that falls through to the policy value.

Also fixed a regression the rebase surfaced: this PR's own new `escalate_paths` strict validation in `loadRepos()` made a single repo's malformed `escalate_paths` throw for the *entire* registry, breaking every other repo's dispatch eligibility (caught by `dispatch.test.mjs`, which never ran green because the PR sat in draft). A malformed value now degrades to `null` (same as absent) in the shared registry read; the dispatch gate's own dedicated `loadRepoEscalatePaths()` still hard-fails malformed `escalate_paths` for that repo's own dispatch decision, unchanged.

Rebased onto latest `develop` (through #636/WM-689, resolving the resulting `Projects.tsx`/`vite.config.ts` conflicts), prettier/eslint clean, `bun test event-runtime/lib` 1048 pass / 0 fail, `Projects.test.tsx` 26/26 pass, `tsc --noEmit` clean, `bun run build` green under develop's current 625 kB entry-chunk budget.

PR #618, opened 2026-08-18
fix(web): return Tab out of a hover card to the trigger (WM-752)

hdkiller:
## Summary

The hover-card panel is `createPortal(..., document.body)`, which puts its tab stops after everything else in the page. Tab from the last in-card control therefore walked out of the document instead of continuing near the trigger, and Shift+Tab from the first control did the same in reverse.

A non-modal `role="dialog"` needs no focus trap, so this is not one: a keydown handler on the panel intercepts Tab **only at the card's edges** and hands focus back to the trigger (the element that was focused when the card opened, falling back to the wrapper). The next Tab then continues from the trigger's own place in the page. Tab between two in-card controls is left entirely to the browser.

A card with nothing focusable holds focus on the panel itself, so both directions are an edge there. Forward Tab from the panel root when the card *does* have controls is not intercepted — DOM order already walks into the card.

Focus landing on the trigger is not focus leaving the card, so the deferred close does not fire behind it; the card stays open until the operator tabs on.

Scope: Tab-out only. The ARIA wrapper ([WM-741](https://linear.app/watt-mind/issue/WM-741)) and the ArrowDown leak ([WM-751](https://linear.app/watt-mind/issue/WM-751)) are separate tickets on the same files.

## Test plan

- [x] `cd event-runtime/web && bun test src/components/HoverCard.test.tsx` — 31 pass / 0 fail (was 27).
- [x] Negative first: the three new behavioural tests (Tab from last control, Shift+Tab from first control, Tab from a control-less card) were observed failing on the pre-fix tree. The fourth new test is a guard that Tab *between* two in-card controls is not intercepted, and passes either way by design.
- [x] Consumers unaffected: `AgentHoverCard.test.tsx` + `Agents.test.tsx` — 21 pass.
- [x] `bun run lint` — 0 errors, no new warnings in the changed files.
- [x] `factory security` — gitleaks, semgrep, actionlint all pass.

UX critique: skipped — keyboard focus management with no visual surface, and the critic was blocked on WM-726.

Fixes WM-752

### DIFF
--- a/event-runtime/web/src/components/HoverCard.test.tsx
+++ b/event-runtime/web/src/components/HoverCard.test.tsx
@@ -3,6 +3,7 @@ import { afterEach, describe, expect, mock, test } from "bun:test";
 import {
   act,
   cleanup,
+  createEvent,
   fireEvent,
   render,
   waitFor,
@@ -45,6 +46,8 @@ function Fixture(props: {
   closeDelayMs?: number;
   onOpenChange?: (open: boolean) => void;
   interactiveTrigger?: boolean;
+  /** Two in-card controls, so the first and last tab stops are distinct. */
+  secondAction?: boolean;
 }) {
   return (
     <HoverCard
@@ -67,6 +70,7 @@ function Fixture(props: {
           <button type="button" onClick={close}>
             Open in Agents
           </button>
+          {props.secondAction ? <button type="button">Copy id</button> : null}
         </>
       )}
     </HoverCard>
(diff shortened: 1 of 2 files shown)


hdkiller:
factory-merge-fix round=1 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=b11482acb7a159e2112beb5bf0400ebccc653684 new=7b894c8973b228d118a24221e6a4d89e7fd2a266

PR #619, opened 2026-08-18
feat(event-runtime): add inbox actions by kind

hdkiller:
## Summary
- add kind-specific Inbox chrome for Replay, Rerun CI, watched Ship, escalation jumps, and diagnose-first resolution
- correlate CI/Ship proposals back to their inbox items and prevent duplicate action submissions
- add bounded fail-open Linear polling plus CI-success and completed-ship auto-resolution
- preserve generic DecisionCard effects for proposal, requeue, and answer flows

## Verification
- `bun test event-runtime/lib event-runtime/web` — 1923 pass, 9 skipped, 0 fail
- `cd event-runtime/web && bunx tsc --noEmit` — pass

## UX
UX critique: required — NOT-ASSESSED

Round 1 returned FIX-FIRST with browser evidence at `http://127.0.0.1:7787/#/inbox` and `/tmp/WM-287-smoke-red-actions.png`; all in-scope findings were addressed (PR chips, post-submit proposal state, and action priority). Round 2 drove the updated live app but exceeded its critic turn budget before returning a verdict, so this PR remains draft rather than treating incomplete review as approval.

Fixes WM-287

### DIFF
--- a/event-runtime/lib/api.mjs
+++ b/event-runtime/lib/api.mjs
@@ -42,7 +42,12 @@ import {
   webhookSecret,
 } from "./config.mjs";
 import { githubWebhookSecret } from "./intake.mjs";
-import { decideInboxItem, getInboxItem, retryInboxDecision } from "./inbox.mjs";
+import {
+  bindInboxProposal,
+  decideInboxItem,
+  getInboxItem,
+  retryInboxDecision,
+} from "./inbox.mjs";
 import {
   applyDecisionEffect,
   decisionEffectPlanner,
@@ -283,6 +288,18 @@ export function createApi({
           );
           if (!schedule) return send(status, body);
           const repo = registry.schedules?.[loop]?.payload?.repo;
+          if (
+            loop.startsWith("ship-") &&
+            typeof repo === "string" &&
+            repo !== "" &&
+            typeof body?.proposalId === "string"
+          ) {
+            bindInboxProposal(db, {
+              kind: "RC READY",
+              repo,
+              proposalId: body.proposalId,
+            });
+          }
           return send(status, {
             ...body,
             schedule: {
--- a/event-runtime/lib/api.test.mjs
+++ b/event-runtime/lib/api.test.mjs
@@ -308,6 +308,43 @@ describe("schedule trigger metadata (WM-259)", () => {
       s.close();
     }
   });
+
+  test("Ship uses the existing schedule route and correlates its proposal to RC READY", async () => {
+    const s = await makeServer({
+      now: () => Date.parse("2026-08-18T12:00:00Z"),
+    });
+    try {
+      createInboxItem(
+        s.db,
+        {
+          kind: "RC READY",
+          title: "bj29 is ready to ship",
+          refs: { repo: "bj29" },
+        },
+        { id: "rc-ready-api" },
+      );
+      const response = await fetch(s.url("/schedules/ship-bj29/run"), {
+        method: "POST",
+        headers: { "content-type": "application/json" },
+        body: "{}",
+      });
+      expect(response.status).toBe(200);
+      const body = await response.json();
+      expect(body).toMatchObject({ loop: "ship-bj29", admitted: true });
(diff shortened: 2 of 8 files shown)


hdkiller:
## Merge review: FIX

Three blocking defects, plus CI has not concluded (required `Verify` check has not started). Escalation gate itself is intact — no new merge/bypass path from the inbox was added — so this is FIX, not ESCALATE.

**1. blocking — `event-runtime/lib/inbox.mjs:754-763` (`linearIssueResolved`): ESCALATED items auto-resolve when the label was never present, not just when it's removed.**
```js
if (kind === "ESCALATED") {
  const labels = (issue?.labels?.nodes ?? []).map((label) => label.name);
  return issue?.state?.type === "completed" || !labels.includes("ai:escalated");
}
```
`ESCALATED` inbox items are created by `createRefusalInboxItem` (`worker.mjs:516`) on any `needs_human` refusal; nothing applies `ai:escalated` to the Linear issue at that moment (the merge-review flow applies it later). So for a worker-refusal escalation on an unlabeled issue, "label absent" reads as "cleared" and the item auto-resolves (`auto:linear_escalation_cleared`) within 60s, discarding the decision request before a human sees it. Fix: resolve on an observed label *transition* (stamp when the label was first seen present, only resolve after that), not on mere absence.

**2. blocking — `event-runtime/web/src/components/InboxActions.tsx:90` (`ciRerunEnvelope`): PR number used as a GitHub Actions run id fallback.**
```js
const runId = original.runId ?? item.refs.pr?.match(/\d+/)?.[0];
```
A `CI RED` item with `refs = {repo, pr}` (no referenced `github.workflow-run.failed` event) yields `runId` = the PR number, which validates and creates a proposal that only fails at approval time with `gh: could not find run <PR#>`. The existing test blesses this (asserts only `payload.repo`). Fix: drop the `refs.pr` fallback; require the actual workflow-run reference or throw the existing "Rerun CI needs the failed workflow run reference" error.

**3. blocking — `event-runtime/lib/inbox.mjs:9`: importing `gql` from the orchestrator CLI can kill the serve process.**
`gql` → `getApiKey()` (`orchestrator/reaper.mjs:133-142`) does `process.exit(1)` when `LINEAR_API_KEY` is missing. This poll runs inside the serve tick (`serve.mjs:141`); a missing/misconfigured key terminates the whole event runtime with one stderr line — `.catch()` can't catch `process.exit`. Every other Linear touchpoint in event-runtime goes out-of-process through `tools/linear.mjs` for exactly this reason. Fix: shell out via `tools/linear.mjs raw <query>` (like `decision-effects.mjs`) or inline a `fetch` that throws instead of exiting on a missing key.

Minor/file-to-Triage (not blocking, noting for awareness):
- `inbox.mjs:809-820` — the `else if` chain order makes the BLOCKED half of the Linear poller effectively unreachable given how `notify.mjs` actually populates `refs` (event refs, not `refs.issue`) for BLOCKED items.
- `reconcileInbox` in `serve.mjs:141` is called without `await`, so poll resolutions/errors are invisible in production.
- `fetchLinearInboxIssues` has no batch cap and a single bad issue id fails the whole query every tick with no partial resolution.
- Ack/Resolve silently disappeared for the new action kinds (human_needed/CI RED/RC READY/ESCALATED) with no feedback on partial bulk-skip — may be intentional per "exactly its actions and no others," flagging for a ticket-level call rather than a fix-in-branch.
- `laterCiSuccess` repo matching assumes short name; a slug-form `refs.repo` won't auto-resolve.
- PR body still says "UX critique: NOT-ASSESSED / remains draft" but PR is not a draft and the ticket has a later "Round 2: SHIP" comment — please sync the body before merge (also missing `run:<id>`).

CI: required `Verify` check has not started yet (queued behind `Fast unit tests`); `Shadow runner fleet available` is green. Branch is MERGEABLE, 2 ahead / 8 behind develop with no overlap on the diverged files. Owned Paths clean, escalate_paths clears (factory `escalate_paths: []`, no merge/bypass control added).

PR #624, opened 2026-08-18
feat(web): config-driven Linear ticket auto-linking and TicketHoverCard (WM-701)

hdkiller:
## What

Ticket ids that appear in payload columns, event envelopes, and agent summaries were plain strings. They now link to `#/tickets/:id` and carry a hover card, and the vocabulary of what counts as a ticket comes from this factory's own configuration rather than a hard-coded list.

- **`components/TicketHoverCard.tsx`** (new) — `ticketTeamsFrom()` reads the team keys off `api.repos()`; `buildTicketPattern()` turns them into `\b(?:CLNT|OPS|WM)-\d{1,6}\b`; `splitTicketRefs()` cuts free text into plain and ticket segments; `TicketText` renders that with hover-card links; `TicketHoverCard` shows the id, state badge, title, latest run and latest PR, and jumps to both the in-app journey and Linear. Built on the WM-700 `HoverCard` primitive, so it opens on focus as well as hover and answers Escape.
- **`components/CustomCell.tsx`** — supports the `x-ui: { kind: "ticket" }` schema annotation (`readCellUi`, `formatCellValue(value, ui)`), and linkifies embedded ids in free-text cells.
- **`views/Ticket.tsx`** — an unindexed id now renders "Unknown or external ticket" with a direct **Open in Linear ↗** link instead of a journey full of dashes; the blocking reason and next action lines linkify the ticket they name.

## Why these two mechanisms and not one

The word-boundary scan is deliberately narrow, because it runs over prose: `UTF-8`, `SHA-256`, and `feat/WM-701-slug` must not become links, and a seven-digit number is not a ticket. That narrowness means it will skip a real ticket from a team this factory does not dispatch — which is exactly what the `x-ui` annotation is for. A column that declares it holds a ticket is believed regardless of team. The tests pin that distinction: removing the annotation branch fails only the test that uses an unconfigured team, which is how I know the branch earns its place.

## Verification

`cd event-runtime/web && bun test src/components/CustomCell.test.tsx src/components/TicketHoverCard.test.tsx src/views/Ticket.test.tsx` — **41 pass, 0 fail**.

Also: full web suite 980 pass / 2 fail, `bunx tsc --noEmit` clean, `bun run lint` 0 errors (620 warnings, under the 621 ratchet), `factory security` pass, prettier clean.

The two full-suite failures are in `Proposals.test.tsx` and are **pre-existing on develop** — I reverted this branch's files to `HEAD` and reproduced the identical two failures without any of these changes. Filed as WM-763.

## Negative testing

Each new behaviour was observed failing before it existed, and then re-checked by breaking it deliberately:

- Dropping `\b` and the digit cap from the pattern → the `UTF-8` / `SHA-256` / `xWM-12` / seven-digit test fails.
- Removing `enabled: open` from the hover card's query → the "does not fetch until opened" test fails.
- Removing the `x-ui` branch from `CustomCell` → the unconfigured-team test fails. The first version of that test used a configured team and passed with the branch deleted; it was rewritten rather than kept.

## Risks

The pattern hook subscribes every custom cell to the shared `["repos"]` query (`notifyOnChangeProps: ["data"]`, `staleTime` 5m, one cache entry app-wide) — worth a look if a wide table feels slower. The hover card and the ticket view now share one `fetchTicketJourney` and one query key, so the fetcher living in a component file is intentional: views import components, not the reverse.

UX critique: blocked — WM-726 / WM-730.

Fixes WM-701

### DIFF
--- a/event-runtime/web/src/components/CustomCell.test.tsx
+++ b/event-runtime/web/src/components/CustomCell.test.tsx
@@ -1,0 +1,255 @@
+import "../test-dom";
+import { afterEach, describe, expect, mock, test } from "bun:test";
+import { cleanup, fireEvent, waitFor } from "@testing-library/react";
+import { CustomCell, formatCellValue, readCellUi } from "./CustomCell";
+import { renderWithClient, restoreApi, withApi } from "../test-render";
+import type { RepoItem } from "../types";
+
+const realFetch = globalThis.fetch;
+
+afterEach(() => {
+  cleanup();
+  restoreApi();
+  globalThis.fetch = realFetch;
+});
+
+function repo(name: string, team: string | null): RepoItem {
+  return {
+    name,
+    path: `/tmp/${name}`,
+    github: `watt-mind/${name}`,
+    team,
+    project: null,
+    base: "develop",
+    deployBranch: null,
+    reportOnly: false,
+    maxInFlight: null,
+    worktreeRoot: null,
+    hasWorktreeUp: false,
+    hasWorktreeDown: false,
+    hasWorktreeWarm: false,
+    verify: null,
+  };
+}
+
+const reposApi = () => ({
+  repos: mock(async () => ({
+    repos: [repo("factory", "WM"), repo("bj29", "CLNT")],
+  })),
+});
+
+/** A cell only renders inside a row; hydration warns otherwise. */
+function renderCell(cell: React.ReactElement) {
+  return renderWithClient(
+    <table>
+      <tbody>
+        <tr>{cell}</tr>
+      </tbody>
+    </table>,
+  );
+}
+
+describe("formatCellValue", () => {
+  test("formats primitives, arrays and objects as before", () => {
+    expect(formatCellValue(undefined)).toMatchObject({
+      text: "—",
+      isComplex: false,
+    });
+    expect(formatCellValue(true)).toMatchObject({
+      text: "true",
+      isComplex: false,
+    });
+    expect(formatCellValue(7)).toMatchObject({ text: "7", isComplex: false });
+    expect(formatCellValue("hi")).toMatchObject({
+      text: "hi",
+      isComplex: false,
+    });
+    expect(formatCellValue([1, 2, 3])).toMatchObject({
+      text: "[3]",
+      isComplex: true,
+    });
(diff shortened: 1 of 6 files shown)


hdkiller:
factory-merge-fix round=1 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=398ed0bf51f69e903616cc7fecdf476a80ae3c85 new=7c22fa058fb86e8844c8e3584c13f7daed292d43

hdkiller:
## CI failure: real build-budget regression, not a flake

`Full verification` → `Typecheck and build event-runtime/web` fails deterministically (not a lane-congestion flake — 27s normal-duration run, no timeout):

```
(!) event-runtime/web/src/subjectJourney.ts is dynamically imported by .../src/App.tsx
    but also statically imported by .../src/components/TicketHoverCard.tsx,
    .../src/views/Ticket.tsx, dynamic import will not move module into another chunk.

✗ Build failed in 5.89s
error during build:
[entry-chunk-budget] Entry chunk budget exceeded: assets/index-Cj0_0NGo.js is 621.20 kB (budget 600 kB).
```

`TicketHoverCard.tsx:24` and `Ticket.tsx:19-27` both statically `import { prNumbersIn, ... } from "../subjectJourney"`, while `App.tsx` only imports that module dynamically for code-splitting. The static import drags `subjectJourney.ts` into the main entry chunk instead of the lazy chunk, pushing the bundle from under 600 kB to 621.20 kB.

Fix: either avoid the static import (pull just the specific pure helper you need out of `subjectJourney.ts`, or restructure so `TicketHoverCard`/`Ticket` don't need the whole module App.tsx lazy-loads), or — if this feature's growth is genuinely expected — re-baseline `ENTRY_CHUNK_BUDGET_BYTES` in `event-runtime/web/vite.config.ts` in its own reviewable commit, per the build tool's own guidance in the error message. Given the rest of the diff reviewed clean (regex safety, HoverCard reuse, security, Owned Paths all verified), this is a build-config/code-splitting gap rather than a logic bug, but it does block merge until resolved.

CI: `Shadow runner fleet available` is green; `Verify` fails deterministically on this build step, confirmed on a rerun (same failure both times) — this is not lane-congestion noise.

PR #628, opened 2026-08-18
fix(tests): clean tracked temporary directories

hdkiller:
## Summary
- add a file-scoped tracked temp-directory helper and migrate event-runtime tests away from direct `mkdtempSync` calls
- add a regression guard plus CI/E2E start/end age-gated temp sweeps
- extend the live Chrome sweep to reap stale Factory-owned temp prefixes while preserving active profiles

## Verification
- `bun test event-runtime/lib event-runtime/cli && actionlint .github/workflows/ci.yml .github/workflows/e2e.yml` — 1,071 pass, 9 environment skips, 0 fail
- `/tmp` before/after matching-prefix snapshot — 1 new entry (target ≤5)
- `bun run lint --max-warnings=621` — pass (611 existing warnings)
- `bun test orchestrator/chrome-sweep.test.mjs` — 6 pass

UX critique: skipped — test, CI, and janitor infrastructure only; no user-facing flow

Fixes WM-760

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,6 +94,11 @@ jobs:
     runs-on: [self-hosted, shadow]
     timeout-minutes: 15
     steps:
+      - name: Sweep stale shared-host temp directories
+        shell: bash
+        run: |
+          find /tmp -maxdepth 1 -mindepth 1 \( -name 'evrt-*' -o -name 'wm334-real-*' -o -name 'pi-chrome-devtools-*' -o -name 'puppeteer_dev_chrome_profile-*' -o -name 'factory-wt-*' \) -mmin +30 -exec rm -rf {} +
+
       - name: Checkout (no action download)
         shell: bash
         env:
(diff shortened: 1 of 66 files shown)


hdkiller:
@/tmp/claude-1000/-home-hdkiller-Develop-factory/ac58ee98-719e-4fef-a3be-30151cc052a1/scratchpad/pr628-comment.md

hdkiller:
Fixed both blocking findings from review:

1. **CI sweep failure guard** — appended `|| true` to all six sweep `find` steps in `ci.yml`/`e2e.yml` (start-of-job and `always()` end-of-job, in `fast-unit-tests`, `verify`, and `browser-e2e`), so a concurrent shared-host race can never red the job it protects.
2. **`commandFixture()` leak** — the helper now forces every fixture root under the `evrt-` prefix (matched by the CI sweep and `chrome-sweep.mjs` janitor) regardless of the caller-supplied prefix, fixing all `merge.test.mjs`/`merge-apply.test.mjs` call sites at the root instead of patching each one. Added a regression test in `tmp-hygiene.test.mjs` asserting `commandFixture` roots match a sweep prefix.

Also fixed a pre-existing bug the rebase surfaced: `verify.test.mjs` and `worker.test.mjs` each had a few `mkdtempSync` call sites the original codemod missed migrating to `tmpDir()` (imports were removed but the call sites weren't updated), which crashed `bun test event-runtime/lib`. Migrated those too.

Rebased onto latest `develop` (through #636/WM-689), prettier/eslint clean, `bun test event-runtime/lib` 1049 pass / 0 fail, `actionlint` clean.

PR #639, opened 2026-08-18
fix(web): re-baseline entry chunk budget to 625 kB (trunk red after merge wave)

hdkiller:
develop is red on `Typecheck and build event-runtime/web`: entry chunk 602.50 kB > 600 kB after #570/#553/#613/#627/#580/#614/#575 landed within 15 minutes (each green alone). Re-baseline to 625 kB (slack 3.6%) and require the next eager-path growth to split instead — tracked in WM-805 (also unblocks #624/WM-701 once split). Vendor split unchanged (xyflow 197 kB, elk own chunk). `bun run build` passes locally.


hdkiller:
Redundant: #560 (WM-291) already re-baselined the entry-chunk budget to 625 kB on develop. The split work stays tracked in WM-805.

PR #664, opened 2026-08-18
fix(inbox): retry empty replan ids and update retarget titles (WM-783, WM-784)

hdkiller:
## Summary
- Treat `applied` + `replanned_awaiting_approval` with empty/missing `newProposalId` as a failed effect so the inbox item stays retryable and retry invokes the decision effect instead of returning `already_applied` (WM-783).
- When retargeting or superseding an unresolved inbox item, update the list title to the replacement proposal so open-list copy matches the new question (WM-784).

## Test plan
- [x] `bun test event-runtime/lib/inbox.test.mjs` — 21 pass after each commit and at the end
- [x] Negative tests observed failing before each fix
- [x] UX critique skipped — backend/inbox ledger, no user-facing web flow in this diff

Fixes WM-783
Fixes WM-784

### DIFF
--- a/event-runtime/lib/inbox.mjs
+++ b/event-runtime/lib/inbox.mjs
@@ -569,16 +569,27 @@ function settleInboxDecision(
   effect,
   { now, recordedEffect = effect, artifactStore },
 ) {
-  const answer = { ...response, effect: recordedEffect };
   const replanned =
     effect.outcome === "applied" && effect.detail === REPLANNED_DETAIL;
-  if (
-    replanned &&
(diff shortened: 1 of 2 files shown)


hdkiller:
factory-merge-fix round=1 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=49c8dd7ab222242e3d9f4703f87106432688f8bb new=6b947634ecb7dba789b1c214c41f7f83e5acc904

PR #707, opened 2026-08-19
feat(extensions): contribute harness content from the extension manifest (WM-849)

hdkiller:
## Summary
- Add `contributes.harness = { floor?, commands?, skills?, subagents? }` to the extension manifest so skills, slash commands, subagents, and the AGENTS.md floor are pack content rather than a `shared/`-only emit pipeline.
- Make `shared/` the built-in `factory/core` harness pack (`shared/factory-extension.json`); `build/emit.mjs` emits it first so `plugins/core/**` and `dist/**` stay byte-identical, then emits every other loaded harness root under its `publisher-extension` namespace.
- Collision: a later pack may not reuse `factory/core`'s name or another pack's plugin slug; emit records an anomaly and still writes the rest.

Fixes WM-849

UX critique: skipped — backend/schema/docs; no user-completable flow.

## Test plan
- [x] `bun build/emit.mjs && bun run check` — 90 generated files match; `git diff --quiet -- plugins dist`
- [x] `bun test event-runtime/lib/extensions.test.mjs` — 33 pass, including WM-849 harness cases
- [x] `bun run lint` — 0 errors
- [ ] Reviewer: glance at `collectHarnessRoots` collision rules and that core dest paths are unchanged

Made with [Cursor](https://cursor.com)

### DIFF
--- a/build/emit.mjs
+++ b/build/emit.mjs
@@ -1,6 +1,6 @@
 #!/usr/bin/env bun
 /**
- * shared/ -> per-harness packaging.
+ * Harness packs -> per-harness packaging (WM-849).
  *
  *   bun build/emit.mjs           # regenerate plugins/ and dist/
  *   bun build/emit.mjs --check   # CI: fail if the tree isn't reproducible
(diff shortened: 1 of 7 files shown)


hdkiller:
factory-merge-fix round=1 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=c6b466d4042382f54bf09143228bc10cf20dadd3 new=6f410d7851209b67837c5042aea9472ced3c3e33

hdkiller:
factory-merge-fix round=2 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=6f410d7851209b67837c5042aea9472ced3c3e33 new=cf6626db1bf500ed8da7701d2146fec89d33a333

PR #748, opened 2026-08-20
feat(extensions): make contributes.connectors a real runtime seam (WM-919)

hdkiller:
## Summary
- `contributes.connectors` is no longer reserved: the loader contract-checks connector modules at load, `serve` starts them after the registry (10s timeout + AbortSignal) and stops them on shutdown, and a failed `start()` is a per-connector configuration anomaly that leaves the rest of the extension loaded.
- Narrow loopback client (`inject`, `inbox.list/get/decide/subscribe`, `proposals.get`, `runs.get`) stamps `source: connector:<ext>/<name>` and `decidedBy: connector:<ext>/<name>:<actor>`; `ctx.secrets` comes from `FACTORY_EXT_*` / `format: "secret"`, never `policy.yaml`.
- Fixture `test-support/extensions/sample/connectors/echo.mjs` plus tests. HTTP/CLI/inbox write-path wiring that sat outside Owned Paths is filed as [WM-942](https://linear.app/watt-mind/issue/WM-942).

Fixes WM-919

UX critique: skipped — no user-facing flow (backend connector lifecycle; `/status.connectors` projection is a helper until WM-942 wires `status-view.mjs`).

## Test plan
- [x] `bun test --timeout 20000 --max-concurrency=4 event-runtime/lib/extensions.test.mjs event-runtime/lib/connectors.test.mjs event-runtime/lib/api-status.test.mjs`
- [x] `bun run check && bun run lint`
- [x] `bun test event-runtime/lib --timeout 20000` (1365 pass)
- [ ] Confirm WM-942 is the follow-up for GET /status, doctor, extensions list column, inbox subscribe write-path, and `docs/event-runtime.md`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -26,6 +26,10 @@ Today an extension can contribute:
 - **hooks** — decision-returning modules run at a named point of the runtime,
   today `approve.before` (immediately before each chain auto-approval); a
   fail-closed waterfall whose every decision is persisted (§Hooks below);
+- **connectors** — long-running processes the runtime starts with `serve`
+  and stops on shutdown; they talk to the outside world (Buzz, Telegram, a
+  tracker) through a narrow loopback client and never hold a DB handle
+  (§Connectors below);
 - **panels** — declarative `factory.panel-view/v1` Overview tiles
   ([`event-runtime-artifact-views.md` §2.6](event-runtime-artifact-views.md#26-panels--factorypanel-viewv1-wm-840));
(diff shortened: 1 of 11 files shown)


hdkiller:
factory-merge-fix round=1 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=864cb8c56b6ba034e3fe2fdb399a10bbaab8a383 new=bf270ad84810509c6c63dcd6c1f39c79b32b5574

PR #914, opened 2026-08-22
fix(control-plane): factory init --control-plane github no longer rewrites the global kind

### LINKED ISSUE #822: fix(control-plane): factory init --control-plane github rewrites the GLOBAL kind, moving every repo
## Problem & Context

Hit during the WM-1006 cutover, 2026-08-22.

`factory init --control-plane github --repo <r>` writes:

```yaml
controlPlane:
  kind: github          # <-- GLOBAL
  github: { repo: ..., teams: {...}, project: Factory, statusField: Status }
```

`kind` is the **workspace-wide default** for every repo that does not set `control_plane:` in `config/repos.yaml`. On the production runner that is **nine repos**, including three `CLNT` client repos. Running the documented init command against that root would have silently moved every one of them off Linear.

The command predates per-repo selection (WM-1007) and was never updated for it. The two are in direct conflict: init assumes a single-repo workspace, while the point of WM-1007 is that a workspace has many.

I worked around it by hand-writing only the `controlPlane.github` sub-stanza and leaving `kind: linear`. Both `config/policy.yaml` files now carry a comment warning the next person not to run init against that root — a comment is not a safeguard.

`docs/quickstart.md` still tells operators to run exactly this command.

## Acceptance Criteria

- [ ] `factory init --control-plane github --repo <r>` writes the `controlPlane.github` stanza and does NOT change `controlPlane.kind` when `config/repos.yaml` already contains more than one repo entry.
- [ ] When `--repo`/`--team` maps to a known repos.yaml entry, init sets `control_plane: github` on **that entry**, not the global default.
- [ ] Setting the global `kind` requires an explicit opt-in flag (e.g. `--global`), and init prints which repos that would move, by name, before writing.
- [ ] Running init against a multi-repo workspace without the opt-in leaves every other repo resolving exactly as before. A test asserts this against a fixture with 3+ repos, one being moved.
- [ ] The regression is observed failing against current code.
- [ ] `docs/quickstart.md` reflects the corrected behaviour; the warning comments in `config/policy.yaml` come out once the tool is safe.

## Source File Pointers

* `lib/control-plane/github.mjs` — `runGithubInit()`, `writeGithubControlPlanePolicy()`, `githubPolicyStanza()`
* `lib/control-plane/index.mjs` — `controlPlaneKindFromPolicy()` / `controlPlaneKindFromRepo()`
* `event-runtime/lib/repos.mjs` — `control_plane` on a repo entry
* `lib/control-plane/github.test.mjs` — existing init tests
* `docs/quickstart.md` — the instruction that is unsafe on a multi-repo root

## Owned Paths

- docs/quickstart.md

## Verification Command

```
bun test --timeout 20000 --max-concurrency=4 lib/control-plane event-runtime/lib/control-plane.test.mjs && bun run format:check && bun run lint
```

hdkiller:
Fixes watt-mind/factory#822

## Summary
- `factory init --control-plane github --repo <r>` no longer overwrites the workspace-wide `controlPlane.kind` default in `config/policy.yaml` when `config/repos.yaml` has more than one repo entry — it now sets `control_plane: github` on just the matching entry instead.
- Setting the global `kind` requires an explicit `--global` flag; init prints which repos would move (by name) before writing.
- A root with 0-1 repos.yaml entries keeps today's behavior (kind is still set) since there is nothing else to move.
- `docs/quickstart.md` documents the safe per-repo default and the `--global` opt-in.

run:$FACTORY_RUN_ID


hdkiller:
factory-merge-fix round=1 finding=fce230ecb528f191483277c52c59c5f19ce934084a2f990e4ed88b5a9d213ee4 old=5f125750a4af947f52ac9c6da2b5f3c83c4339b7 new=22e4eeb62feca51c7cb71d1c232670513617c9bb

PR #936, opened 2026-08-22
docs: pamcode-form orchestrator guide

### LINKED ISSUE #925: docs: pamcode-form rewrite of the orchestrator guide (docs/orchestrator-pamcode.md)
Evaluate rstacruz/pamcode (pseudocode-in-Markdown grammar for agent workflows) as a format for the factory's operator-facing guides.

Deliverable: `docs/orchestrator-pamcode.md` — the content of `docs/orchestrator.md` re-expressed in pamcode (Input / Skill dependencies / Workflow `begin()` / `### name()` subroutines / Guidelines), so we can compare readability and token cost against the current prose guide.

Acceptance criteria:
- `docs/orchestrator-pamcode.md` exists and follows the pamcode spec sections and constructs.
- All five orchestration loops, the escalation table, the trap catalog, and the delegation rule from `docs/orchestrator.md` are represented (as pseudocode or prose, whichever is clearer).
- No behavioural claim contradicts `docs/orchestrator.md`.

Owned Paths: `docs/orchestrator-pamcode.md`

Verification: manual read.

Source: https://github.com/rstacruz/pamcode

hdkiller:
Fixes #925

Adds `docs/orchestrator-pamcode.md` — `docs/orchestrator.md` re-expressed in [pamcode](https://github.com/rstacruz/pamcode), a pseudocode-in-Markdown grammar for agent workflows. Purpose is to compare the two formats on readability and token cost before deciding whether any operator-facing guide should adopt it.

## What changed in the shape

- The six numbered loops collapse into one `begin()` that is the actual shift loop — observe → land → recover → refill → dispatch → refresh → report. Ordering that the prose guide never states (land before dispatch, because merging is what frees slots) becomes the document.
- Each loop is a `def`: `assess()`, `refill-supply()`, `dispatch()` / `redispatch()`, `land-prs()`, `watchdog-sweep()`, `refresh-stack()` / `clear-tree()`, `escalate()`.
- Escalation triggers stop being a separate section — they are a `continue` at the top of `land-prs()`, before the reviewer is even spawned.
- The four pre-merge gates read as sequential guards with early exits rather than a checklist satisfiable out of order.
- Delegation gets a syntax: `subagent(factory-merge-reviewer) { … }` puts the boundary where the ~30s rule is checkable.

## What deliberately stayed prose

The trap catalog, the notify table, the delegation rule, and the command reference are constraints, not control flow — pseudocode would only obscure them. The rationale lines ("being wrong-but-reversible while moving beats being right-but-stopped", why a dirty checkout freezes the factory) are kept verbatim in `## Guidelines` and above the relevant `def`s. Pamcode is a good fit for the parts of this doc that are a procedure and a bad fit for the parts that are judgment; the spec itself recommends preferring prose for intent.

419 lines vs 423, but ~25% fewer words (2599 vs 3458).

## Handoff

- **Scope:** one new file, `docs/orchestrator-pamcode.md`. `docs/orchestrator.md` is untouched and is marked normative in the new file's header — where the two disagree, the original wins.
- **Verification:** docs-only, no code paths touched. Manual read against `docs/orchestrator.md` section by section; every loop, the escalation table, the trap catalog, and the Loop 6 delegation rule are represented. Command names were checked against the live CLI — the doc uses `factory ticket <verb>` and notes `factory linear` as the deprecated alias (`factory linear` now prints a deprecation notice).
- **UX critique:** N/A — no user-facing surface.

## Discovered while doing this

`bin/worktree-common.sh` calls `realpath -m`, which macOS `realpath` does not support — `bin/worktree-up.sh` prints six `illegal option -- m` errors per run and still succeeds. Filing separately; not fixed here.


hdkiller:
CI note: the 5 planEvent memoPin failures are trunk-level test pollution (env-var restore bug in planner.test.mjs), fixed in #943 — rebase/rerun after it lands.

PR #978, opened 2026-08-23
fix(outbox): retain published rows and index drain

### LINKED ISSUE #953: fix(outbox): no retention and no index — unbounded growth plus a full-table scan on every serve tick
## Problem & Context

The outbox table grows forever and the per-tick drain degrades into a full-table scan:

- No retention exists: a repo-wide search finds **no** `DELETE FROM outbox` anywhere in lib/cli — only `UPDATE outbox SET published_at = ?` (`event-runtime/lib/outbox.mjs:29`). Every completed run permanently adds a row.
- `publishOutbox` runs on every serve tick (cli/serve.mjs) executing `SELECT seq, event_json FROM outbox WHERE published_at IS NULL ORDER BY seq` (`outbox.mjs:22`), and `published_at` has **no index** in any schema version (`event-runtime/lib/db.mjs:104-109`) → full scan of an ever-growing table on every tick. `status-view.mjs` repeats the pattern for the anomaly deck.

On a long-lived factory DB this is an unbounded-growth + quadratic-drain path: ticks get progressively slower, and nothing prunes.

## Acceptance Criteria

- [ ] An index on `(published_at, seq)` (or equivalent covering index for the drain query) is added via the standard schema migration path.
- [ ] Published outbox rows are retained only up to a bounded window (mirror the existing `sweepRuntimeRetention` approach and its configurability); the retention sweep is exercised by a test.
- [ ] Regression test proves the drain no longer scans the full table: e.g. seed 200k published rows + 1 unpublished, assert `publishOutbox` publishes the pending row and that the query plan uses the index (or measure bounded row visits).
- [ ] Retention never deletes unpublished rows; crash between publish and prune loses nothing (at-least-once preserved).
- [ ] Existing `event-runtime/lib/outbox.test.mjs` assertions unchanged.

## Source File Pointers

* `event-runtime/lib/outbox.mjs:21-31` — drain query and publish stamp
* `event-runtime/lib/db.mjs:104-109` — outbox schema (no index)
* `event-runtime/lib/status-view.mjs:202` — second unindexed scan of the same shape
* existing runtime-retention implementation to mirror (search `sweepRuntimeRetention`)
* `event-runtime/lib/outbox.test.mjs`

## Owned Paths

* event-runtime/lib/outbox.mjs
* event-runtime/lib/outbox.test.mjs
* event-runtime/lib/db.mjs
* event-runtime/lib/status-view.mjs

## Verification Command

```
bun test --timeout 20000 event-runtime/lib/outbox.test.mjs event-runtime/lib/db.test.mjs
```

hdkiller:
Fixes #953

UX critique: skipped — backend/database maintenance; no user-facing flow.

### DIFF
--- a/event-runtime/lib/db.mjs
+++ b/event-runtime/lib/db.mjs
@@ -415,6 +415,16 @@ export const MIGRATIONS = [
       `);
     },
   },
+  {
+    version: 13,
+    name: "outbox_publish_drain_index",
+    up(db) {
+      db.exec(`
+        CREATE INDEX IF NOT EXISTS idx_outbox_published_seq
+          ON outbox (published_at, seq);
+      `);
+    },
+  },
 ];
 
 export const CURRENT_SCHEMA_VERSION =
--- a/event-runtime/lib/outbox.mjs
+++ b/event-runtime/lib/outbox.mjs
@@ -9,14 +9,48 @@
  */
 import { tx } from "./db.mjs";
 
+export const DEFAULT_OUTBOX_RETENTION_DAYS = 14;
+const DAY_MS = 24 * 60 * 60 * 1000;
+
+/** Delete published rows older than the configured retention window. */
+export function sweepPublishedOutbox(
+  db,
+  { retentionDays = DEFAULT_OUTBOX_RETENTION_DAYS, now = Date.now() } = {},
+) {
+  if (!Number.isFinite(retentionDays) || retentionDays <= 0) {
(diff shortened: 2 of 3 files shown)


hdkiller:
factory-merge-fix mechanical=format_and_lint finding=20682d08b43619f1ef89214a469175546260733a7a6cb0ea902b6aa0e51b7984 old=671ffbebe1e085511cf97369822d95911ea757d1 new=1544cd908dc735e4b765ff503b39628a62144652

PR #984, opened 2026-08-23
fix(worker): abort force-failed executions

### LINKED ISSUE #954: fix(worker): forceFailRun neither aborts the active execution nor releases the ticket claim — killed runs keep burning their slot
## Problem & Context

`forceFailRun` (`event-runtime/lib/worker.mjs:3632`) and `cancelRun` (`worker.mjs:3619-3625`) are asymmetric. `cancelRun` looks up `ACTIVE_EXECUTIONS` and calls `active.abort(reason)` so the adapter child stops immediately; `forceFailRun` only transitions the DB to FAILED (+ `finishAttempt`) and returns — the running adapter child is left to natural completion, and nothing releases the ticket claim.

Consequences: an operator force-failing a stranded run keeps burning the full remaining agent runtime (minutes to hours of a concurrency slot) on a run they already killed, the worker then unwinds through the late-completion path (its own terminal write loses the race and is swallowed), and the control-plane ticket stays claimed/In Progress until unwind instead of being released promptly.

Verified by code read at origin/develop (76eacb7e): the `ACTIVE_EXECUTIONS` abort block exists in `cancelRun` directly above `forceFailRun`; `forceFailRun` contains no equivalent.

## Acceptance Criteria

- [ ] `forceFailRun` aborts any active execution for the run (`ACTIVE_EXECUTIONS` lookup + `abort(reason)`), matching `cancelRun` semantics.
- [ ] The ticket claim is released (or a release is scheduled as part of unwind) when the force-failed run was holding one — an operator force-fail must not leave a ticket stuck In Progress until agent completion.
- [ ] Regression test: fake adapter sleeping N seconds; `forceFailRun` at t≈0; assert the adapter aborts immediately (not at t=N), the run is FAILED with the journaled reason, and the claim-release path runs. Observed red before the fix.
- [ ] LEASED-state force-fails keep their existing behavior unchanged.
- [ ] Existing worker tests not weakened.

## Source File Pointers

* `event-runtime/lib/worker.mjs:3619-3625` — `cancelRun`'s abort pattern to replicate
* `event-runtime/lib/worker.mjs:3632+` — `forceFailRun`
* `event-runtime/lib/worker.mjs:1976-1999` — `ACTIVE_EXECUTIONS` registry
* `event-runtime/lib/worker.test.mjs`

## Owned Paths

* event-runtime/lib/worker.mjs
* event-runtime/lib/worker.test.mjs

## Verification Command

```
bun test --timeout 20000 event-runtime/lib/worker.test.mjs
```

hdkiller:
Fixes #954

UX critique: skipped — backend worker behavior with no user-facing surface.

### DIFF
--- a/event-runtime/lib/worker.mjs
+++ b/event-runtime/lib/worker.mjs
@@ -2664,12 +2664,15 @@ export async function executeClaimed(
     if (deadlineExpired) outcome = { ...(outcome ?? {}), timedOut: true };
 
     if (abortController.signal.aborted && !deadlineExpired) {
+      const terminalState = db
+        .query(`SELECT state FROM runs WHERE run_id = ?`)
+        .get(runId)?.state;
       if (mayMutateClaimedTicket()) {
         try {
           unclaimTicketFn({
             repo: repoName,
             ticket: ticketId,
-            why: "cancelled",
+            why: terminalState === "FAILED" ? "force_failed" : "cancelled",
             log: null,
           });
         } catch {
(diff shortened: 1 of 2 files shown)


hdkiller:
factory-merge-fix mechanical=format_and_lint finding=20682d08b43619f1ef89214a469175546260733a7a6cb0ea902b6aa0e51b7984 old=7a80f9f3c5050fd3564ef3040b7e6bb1aaafd731 new=425a3043137f11a3315d599941f3336090d4439e

PR #1007, opened 2026-08-23
fix(notify): resolve the transport from policy, not a hardcoded path (WM-879)

vvb-1:
Closes the WM-879 debt that both `bin/factory` and `event-runtime/lib/notify.mjs` name in their own comments.

## What happened

Fresh instance, GitHub-Issues control plane, `factory-merge` running unattended. It reviewed two PRs, merged one, and correctly escalated the other — then this, from inside the escalation path:

```
python3: can't open file '/home/krily/Develop/hdkiller/scripts/notify.py': [Errno 2] No such file or directory
```

Nothing was lost (the escalation is traceable through the PR labels and the ticket comment), but the channel whose entire job is "tell a human this needs you" is the one that failed, at the one moment it mattered. On any machine that is not the original operator's, that fallback leg cannot succeed.

## Change

Two hardcoded paths, one resolution:

- `bin/factory notify` — the direct fallback used when the event runtime is unreachable, or the message is unstructured, now resolves via `lib/notify.mjs` like every other caller (the path WM-795 established). Unconfigured is silent-and-said-so at exit 0, not a crash.
- `event-runtime/lib/notify.mjs` — `DEFAULT_NOTIFY_CMD` is the resolved command or `null`. `notifyPending` skips only the push leg when enabled-but-unconfigured, and logs it once per referent. The durable inbox row is written before that branch, so the human can still see the item in the UI.

The test that asserted the private path now asserts the absence of one, which is the actual invariant.

## Verification

- `bun test event-runtime/lib --timeout 20000` — 1562 pass, 0 fail
- `bun build/emit.mjs --check` — clean
- `factory notify "msg"` and `... | factory notify -` against a stub transport: both deliver, argv and stdin shapes unchanged
- unconfigured: one line on stderr, exit 0

### DIFF
--- a/bin/factory
+++ b/bin/factory
@@ -240,15 +240,20 @@ case "$cmd" in
     fi
 
     # Unreachable runtime (or an unstructured legacy message): preserve the
-    # old argv/stdin transport exactly. WM-795 gave `factory notify`'s callers
-    # (runners/run-agent.sh) a policy-configurable command via lib/notify.mjs,
-    # but this direct fallback still hardcodes the operator's private
-    # notify.py path; WM-879 owns dropping it in favour of the same
-    # config/policy.yaml resolution.
+    # old argv/stdin transport, but resolve the command the same way every
+    # other caller does (lib/notify.mjs — WM-795), rather than hardcoding the
+    # operator's private notify.py path. A clone that has configured no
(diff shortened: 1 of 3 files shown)


vvb-1:
I have read the CLA Document and I hereby sign the CLA

PR #1008, opened 2026-08-23
fix(claude): do not enable an OS sandbox the host cannot start

vvb-1:
## Symptom

A read-only `work-scan@1` run on a fresh Linux instance, twice, ~$0.70 each:

```
Refused with `needs_human`: the Bash tool is non-functional in this environment
(every invocation, even `echo hello`, fails with
`apply-seccomp: write /proc/self/setgroups ... Permission denied`)
```

The agent behaved correctly — it refused rather than inventing data. But the harness works perfectly on that host outside the sandbox, so the refusal was caused by the adapter's own policy, and it burned a full context window plus a human's attention to say so.

## Cause

`buildClaudeSettings` emits, for `mutating: false`:

```js
sandbox: {
  enabled: true,
  autoAllowBashIfSandboxed: true,
  allowUnsandboxedCommands: false,
  ...
}
```

On Linux that sandbox is a nested user namespace. Hosts that forbid one (hardened kernels, containers, LXC/LXD) fail *before* the policy is applied, and `allowUnsandboxedCommands: false` turns "sandbox unavailable" into "no Bash at all". The comment at the top of the file already anticipates this for the guest case ("with `allowUnsandboxedCommands: false`, would deny every Bash call if absent") — it is equally true on an unsupported host.

## Fix

Probe once per process; when unavailable emit `sandbox: { enabled: false }`.

The probe mirrors the failing step exactly rather than a nearby one:

```
unshare --map-root-user --setgroups deny true
```

On the affected host plain `unshare -Ur` **succeeds** and only the setgroups-deny fails, so the weaker probe would have reported a sandbox that dies on first use.

`FACTORY_CLAUDE_OS_SANDBOX=0/1` forces the decision for an operator who knows better.

## Why this does not weaken the boundary

The write boundary never rested on the OS sandbox:

1. the `permissions.deny` entry still refuses Edit under the checkout;
2. the worker's post-run repository integrity gate is what actually proves a read-only run mutated nothing.

The OS sandbox is defence in depth on top of those, so this degrades isolation on hosts that could never provide it anyway — instead of degrading the run to a refusal. It is visible in the emitted policy, so an operator can see which mode a run used.

## Verification

- `bun test event-runtime/lib --timeout 20000` — 1564 pass, 0 fail (two new tests: the disabled branch keeps the deny policy, and the override/probe contract)
- `bun build/emit.mjs --check` — clean
- the same injected `factory.work.requested` run that refused twice now reaches COMPLETED with a valid `factory.work-plan/v1` artifact

### DIFF
--- a/event-runtime/lib/adapters/claude.mjs
+++ b/event-runtime/lib/adapters/claude.mjs
@@ -23,7 +23,7 @@
  * `SANDBOX_DEFERRAL_REASON` below is the whole rationale, and it names what
  * the pi path did not have to solve.
  */
-import { spawn } from "node:child_process";
+import { spawn, spawnSync } from "node:child_process";
 import {
   createWriteStream,
   mkdirSync,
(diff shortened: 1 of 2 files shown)


vvb-1:
I have read the CLA Document and I hereby sign the CLA

PR #1025, opened 2026-08-24
feat(schedules): enable the work-bj29 autonomous clock

hdkiller:
Operator decision: bj29 self-dispatches its BJ29 Coaching queue on a 30m clock like factory. Schedule-sourced, so escalate_paths still parks sensitive-path tickets for operator dispatch. Depends on #1024 (operator escalate_paths bypass) for the sensitive ones.

## Verification
registry.test 46 pass, loop.test 12 pass.


hdkiller:
Closing — wrong layer. Enabling a client repo's clock by editing the tracked (public) schedules.json commits a client name into the open-source repo. The correct home is the instance-local schedule overlay (config/schedule.yaml, gitignored). Re-doing via the overlay; a separate cleanup will move the other client loops out of the kernel file too.

Issue #1069, opened 2026-08-29
feat(cli): factory ask — one read-only surface for factory state (queue, in-flight, held, runs, noop reasons, spend)


hdkiller:
## Problem & Context

Every surface that wants to answer *"what is the factory doing right now"* re-derives it from a different place. `orchestrator/queue.mjs` knows what is dispatchable, `orchestrator/next.mjs` knows what would be picked next, `orchestrator/economics.mjs` parses transcripts for spend, `orchestrator/status.mjs` composes a snapshot, and the reason a planner declined to act lives only in `proposals.reason` inside `runtime.db`. There is no single call that returns factory state as data.

That is fine while the only consumer is a human reading terminal output. It stops being fine the moment a second consumer exists — and three are queued: the `@factory` Buzz agent (needs to answer questions without shelling out to five scripts and parsing their prose), the context graph (needs the same nouns as typed rows), and the web Overview. Building the read surface once and pointing all three at it is strictly cheaper than three parsers over the same prose.

This ticket adds one read-only verb, `factory ask`, that returns a stable JSON document describing current factory state. No mutation, no model, no network beyond the control plane and forge reads the existing modules already do.

Note for the implementer: `bin/factory` dispatches verbs from an explicit `case` statement (`bin/factory:65`), so the verb must be registered there as well as implemented.

## Acceptance Criteria

- [ ] `factory ask --json` returns one document with these top-level keys, each an array or object of typed rows (never pre-formatted strings): `queue` (dispatchable tickets), `inflight` (claimed, with age and last heartbeat), `held` (`ai:blocked` / `ai:escalated`, with the question), `recent` (runs in the last 24h with agent, adapter, model, outcome, cost), `noop` (latest planner decline per event type with its `reason`), `spend` (window totals reusing `lib/spend.mjs`, not a second parser).
- [ ] `factory ask` with no `--json` prints a human summary of the same document; the JSON is the source and the text is rendered from it, so the two cannot disagree.
- [ ] `factory ask --section queue,held` returns only the named sections; an unknown section name exits non-zero naming the valid set.
- [ ] Every section is independently fallible: if the control plane is unreachable, that section carries `{ error: "<reason>" }` and the others still return. A section that fails is never rendered as empty — an unreachable tracker must not read as "no tickets" (`docs/protocol.md` §"GitHub Issues binding").
- [ ] `--repo <name>` scopes to one repo; absent, every non-`report_only` repo in `config/repos.yaml` is included.
[...]

PR #1093, opened 2026-08-29
feat(cli): factory ask — one read-only surface for factory state (#1069)

hdkiller:
## What
Adds `factory ask` — one read-only verb that returns factory state as a stable
JSON document (and a human summary rendered from it): `queue` (dispatchable),
`inflight` (claimed, with age + heartbeat), `held` (`ai:blocked`/`ai:escalated`,
with the question), `recent` (runs in the last 24h with agent/adapter/model/
outcome/cost), `noop` (latest planner decline per event type with its reason),
and `spend` (window totals via `lib/spend.mjs`).

## Why
Every surface that answers "what is the factory doing right now" re-derives it
from a different place. This builds the read surface once so the `@factory`
agent, the context graph and the web Overview parse typed nouns instead of
three parsers over the same prose (#1069).

## Acceptance
- `factory ask --json` returns one document of typed rows (never pre-formatted strings).
- Bare `factory ask` prints a human summary rendered from that same document.
- `--section queue,held` filters; an unknown section exits non-zero naming the valid set.
- Every section is independently fallible: a failing tracker read carries an
  error keyed by repo and is never rendered as an empty queue; a missing ledger
  isolates to `recent`/`noop` while the others still return.
- `--repo <name>` scopes the tracker sections; absent, every non-`report_only`
  repo in `config/repos.yaml` is included. (`recent`/`noop`/`spend` are runtime-
  instance wide, since the ledger and transcript log are not repo-partitioned.)
- Read-only by construction: the module imports no write verb, and a test
  asserts no `claim`/`transition`/`setLabels`/`file`/`comment` reaches a fake
  control plane during a full `ask`.
- Tests: all sections on a seeded fixture, per-section error isolation,
  `--section` filtering, unknown-section exit code, and the read-only assertion
  (15 tests, all pass).

## Owned Paths
- `orchestrator/ask.mjs`
- `orchestrator/ask.test.mjs`
- `bin/factory`

## Notes / deferrals
- The held-ticket `question` is taken from the newest comment via the neutral
  `listComments` verb (adapter-agnostic), rather than Linear-only label-history
  reconstruction — a reasonable first cut that stays read-only and works on
  GitHub Issues too.

### DIFF
--- a/bin/factory
+++ b/bin/factory
@@ -72,6 +72,7 @@ case "$cmd" in
     run_bun tools/ticket.mjs "$@"
     ;;
   queue)        run_bun orchestrator/queue.mjs "$@" ;;
+  ask)          run_bun orchestrator/ask.mjs "$@" ;;
   next)         run_bun orchestrator/next.mjs "$@" ;;
   state)        run_bun orchestrator/state.mjs "$@" ;;
   status)       run_bun orchestrator/status.mjs "$@" ;;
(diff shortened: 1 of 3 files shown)


hdkiller:
Duplicate of #1102 (issue #1069).

Issue #1114, opened 2026-08-29
fix(ask): read-only runtime.db open fails on an uncleanly-shut-down WAL database


hdkiller:
Found while reviewing #1069 (PR #1102). Behaviour is graceful today; filing so it is tracked rather than rediscovered.

## Problem

`orchestrator/ask.mjs` opens `runtime.db` with `new Database(file, { readonly: true })` — deliberately, because `openDb()` migrates and sets pragmas, which a read-only verb must not do (#1069 AC "read-only is enforced by construction").

A read-only SQLite connection cannot create the WAL index (`-shm`). A database left with a `-wal` but no `-shm` — the state after an unclean shutdown or a crash — therefore fails to open, and `factory ask`'s `recent` and `noop` sections report `unavailable` even though the data is intact and readable by any writer.

## Impact

Low and self-correcting: the failure direction is right (it reports `unavailable`, never a silent "no runs"), the sections are independently fallible so the rest of `ask` still answers, and it clears the moment any writer reopens the database. But it is precisely after a crash that someone runs `factory ask` to find out what happened, so the degraded window overlaps the moment of need.

Documented in the `openRuntimeDb` docstring on PR #1102; this issue tracks doing better than documenting it.

## Options

1. On open failure, retry once via an immutable-mode URI (`file:...?immutable=1`) — reads a WAL-less snapshot without writing. Risks a stale read if the `-wal` holds recent commits.
2. Detect the `-wal`-without-`-shm` case and report a specific reason ("runtime database was not shut down cleanly; start the runtime or run any writer once") instead of the raw SQLite error, so the operator gets the fix rather than the symptom.

Option 2 is the honest minimum; option 1 trades correctness for availability and probably should not be taken for a surface whose contract is "never report a failed read as empty".

## Owned Paths

- orchestrator/ask.mjs
- orchestrator/ask.test.mjs

## Verification Command

```
bun test --timeout 20000 orchestrator/ask.test.mjs
```

## AI-Ready Specification (authoritative)

The original diagnosis and impact above remain evidence. This section is authoritative for implementation and resolves the earlier options in favor of the honest specific-error behavior (option 2). Do **not** add an immutable-mode fallback.

### Problem & Context

[...]

Issue #1113, opened 2026-08-29
perf(control-plane): the GitHub adapter's spawnSync transport serializes every read, defeating caller concurrency


hdkiller:
Found while implementing the bounded-concurrency fix for #1069 (PR #1102), and worth filing because it silently defeats a fix that looks correct at the call site.

## Problem

`lib/control-plane/github.mjs:18,170,433` runs every `gh` call through `spawnSync` from `node:child_process`. `spawnSync` blocks the event loop for the whole subprocess lifetime, so `async` methods on the adapter return already-settled promises and **cannot overlap**, no matter what the caller awaits around them.

Measured on the factory repo (98 tickets, 14 held), five `listComments` reads:

| pattern | elapsed |
|---|---|
| `Promise.all` over 5 | 13.0s |
| `for … await` over 5 | 12.5s |

Identical — the fan-out buys nothing. The same five reads against the Linear adapter (`lib/control-plane/linear.mjs` → `gql()` → `await fetch`) do parallelize.

## Impact

Any caller that bounds concurrency to speed up a batch of adapter reads gets the correct-looking code and none of the benefit, on the adapter the factory repo itself uses (`config/repos.yaml` sets `control_plane: github`). `factory ask --section held --repo factory` measured 43s serialized vs 40s at concurrency 5 — within noise, where the same change on Linear would be roughly linear in the limit.

More generally this makes every GitHub-plane read a stop-the-world call: nothing else on the event loop — timers, other repos' reads, the API server if it shares a process — makes progress while `gh` runs.

## Suggested fix

Switch the `exec` seam to an async spawn (`Bun.spawn` / `child_process.spawn` with a promise wrapper) and `await` it. The seam already exists and is injected in tests (`exec = spawnSync` at `github.mjs:433`), so the change is contained: the adapter's methods are already `async`, and `github.test.mjs` passes its own fake `exec`. The fakes would need to return promises.

Worth checking the same pattern in `lib/forge/` before closing.

## Acceptance Criteria

- [ ] The GitHub adapter's `exec` seam is asynchronous (`Bun.spawn` or `child_process.spawn` wrapped in a promise); no `spawnSync` remains in `lib/control-plane/github.mjs`.
- [ ] Two concurrent adapter reads overlap: a test injects a fake `exec` that resolves after a delay and asserts `Promise.all` over N calls completes in ~1x the delay, not ~Nx.
[...]

Issue #1112, opened 2026-08-29
fix(ask): "QUEUE — dispatchable now" overstates — it is the tracker predicate, not what next would start


hdkiller:
Found while reviewing #1069 (PR #1102).

## Problem & Context

`orchestrator/ask.mjs` builds the `queue` section from `cp.listDispatchable()`, which is the correct tracker-side predicate (Todo + `ai:agent-ready` + unassigned + no open blocker, in queue order). But the human header reads:

```
QUEUE — dispatchable now (58)
```

`listDispatchable` does not apply what `orchestrator/next.mjs` applies on top of it: Owned Paths collision against in-flight work, and per-repo capacity. So tickets listed under "dispatchable now" may be ones `next` would decline this tick — which is exactly the gap `factory queue` narrates separately ("nothing startable — all ready tickets collide with running or with each other's unparseable Owned Paths").

Reproduced live: `ask` reported 58 dispatchable against 20 in flight on a repo whose cap is 20 — zero of the 58 were startable at that moment.

Wording, not data — the rows themselves are correct and correctly sourced. But "dispatchable now" invites the reading "these would start if you ran dispatch", which is false whenever capacity is full or paths collide, and the queue-depth number is the one a human uses to decide whether the factory is starved.

Still present on `origin/develop` as of 2026-08-30: `orchestrator/ask.mjs:672`.

## Suggested fix

Rename to `QUEUE — eligible (n)`. Do **not** add the collision/capacity gate: `ask` is deliberately a read surface over the tracker predicate, and duplicating `next.mjs`'s selection logic here would create the second implementation #1069 set out to avoid.

The string appears in exactly two places in `ask.mjs`, both inside `formatAsk`: the header at line 672 and the empty-state string `"no dispatchable tickets"` passed to `renderRows` just below it. A grep of `origin/develop` finds no other copy — no doc, skill or `dist/` surface quotes the header, so the two owned files are the whole closure.

## Acceptance Criteria

- [ ] The human-rendered header reads `QUEUE — eligible (n)` (or an equivalent word that does not claim the rows would start now).
- [ ] Rows, their order, and their source (`cp.listDispatchable()`) are unchanged; no selection logic from `orchestrator/next.mjs` is duplicated into `ask.mjs`; `ask` still imports no write verb (the existing static/dynamic read-only assertions in `ask.test.mjs` must still pass).
[...]

Issue #1111, opened 2026-08-29
fix(spend): todaysSpendBreakdown cannot distinguish "nothing spent" from "log dir unreadable"


hdkiller:
Found while reviewing #1069 (PR #1102), which correctly reuses this parser rather than writing a second one.

## Problem & Context

`lib/spend.mjs` wraps the whole day-scan in a bare `try {} catch {}`:

```js
} catch {
  /* no log dir yet — nothing spent */
}
return { usd: reported + estimated, reported, estimated, runs };
```

Any failure — missing directory, permissions, an unreadable file mid-scan — returns `{ usd: 0, reported: 0, estimated: 0, runs: 0 }`, which is byte-identical to a genuinely idle day.

This is not only a reporting nit. `budgetExhausted()` is built on the same breakdown and is the gate that stops every agent-spawning stage when the day's budget is spent (`orchestrator/tick.mjs`, `orchestrator/queue.mjs`, `orchestrator/next.mjs`, `orchestrator/watchdog.mjs`, `event-runtime/lib/planner.mjs`, `event-runtime/lib/auto-approval.mjs` all consume it). A log directory that cannot be read makes the gate return `null` — "under budget" — and the factory keeps dispatching. The module docstring already argues that "a budget gate that cannot see a third of the runs is worse than no gate"; this is the same argument applied to a gate that can see none of them.

It also propagates to `factory ask --json` (`orchestrator/ask.mjs` sets `doc.spend.today`), where `spend.today.runs === 0` is currently unfalsifiable.

Still present on `origin/develop` as of 2026-08-30 — verified against `lib/spend.mjs`, unchanged since the review that found it.

## Suggested fix

Distinguish the cases rather than collapsing them:

- `ENOENT` on the log directory genuinely is "nothing spent yet" — keep returning zeros, but carry `{ scanned: false, reason: "no log directory" }`.
- Any other error must surface: return an `error` field, and make `budgetExhausted()` fail **closed** (treat an unreadable log dir as "cannot prove we are under budget") rather than open.

`budgetExhausted()` already returns `string | null` and every caller only tests truthiness and prints the string, so failing closed with a reason string is source-compatible; no caller needs to change. Added fields on the breakdown are additive for `ask --json`.

Pre-existing behaviour in the shared parser, not introduced by #1069.

## Acceptance Criteria

[...]

PR #1117, opened 2026-08-29
fix(cli): factory ask — correct in-flight age/heartbeat, qualify held comments, bound comment fan-out (#1069)

### LINKED ISSUE #1069: feat(cli): factory ask — one read-only surface for factory state (queue, in-flight, held, runs, noop reasons, spend)

hdkiller:
Follow-up to #1102 (merged). Addresses the cold review of #1069; #1102 landed while these fixes were in progress, so they ship as their own PR on top of it. Same Owned Paths.

Fixes #1069

## BLOCKING 1 — in-flight age and heartbeat were wrong on every GitHub-bound repo

`lib/control-plane/github.mjs` pins `startedAt: null` and `lastCommentAt: null`, and `config/repos.yaml` puts the factory repo on that adapter — so this was the primary path, not an edge case. A ticket claimed 12 minutes ago rendered as `30d old / heartbeat never`.

- `claimedAt` now follows `startedAt → updatedAt → createdAt`, the same precedence as `reaper.mjs` `lastActivity()`, instead of silently degrading to issue-creation time.
- `heartbeatOf()` returns `{ at, source }` with `heartbeatSource: "comment" | "updatedAt" | "unknown"`, so a consumer can tell a real heartbeat from a proxy. `updatedAt` advances on claims and label swaps, so it is an upper bound on silence, not a heartbeat — the JSON now says which it is.
- The renderer prints `heartbeat 11m ago` / `updated 12m ago` / `heartbeat unknown`. A heartbeat we never read is no longer reported as one that never happened — the same "unavailable is not empty" rule the section level already follows.

## BLOCKING 2 — the fixture was more capable than the adapter it stood in for

`memoryControlPlane.asTicket()` synthesizes `lastCommentAt` from the last comment, and the old fixture set `startedAt` explicitly, so the suite only ever saw Linear's richer ticket and could not see finding 1. Added a GitHub-shaped in-flight ticket (`startedAt: null`, `lastCommentAt: null`, `comments: []`, `updatedAt` 12m ago, `createdAt` 30d ago) and a dedicated regression test asserting age and heartbeat resolve from `updatedAt` — explicitly not `30 * DAY`, explicitly not `never`.

## FIX 3 — `held` could present the human's ANSWER as "the question"

`reply-detection` anchors on the `ai:blocked` label-add and only falls back to newest-comment when there is no add event; the neutral contract exposes no label history (GitHub has none), so this is always the fallback. Once a human replies but before the ticket is unblocked, the newest comment is their answer — wrong in exactly the "what is waiting on me" view.

True parity is not reachable through the contract, so the fix is provenance: rows carry `questionSource: "newest-comment"`, and the renderer says `last comment:` under a `(newest comment shown — may be the reply, not the ask)` header rather than presenting it unqualified.

## FIX 4 — bounded concurrency for the held comment fan-out

`listComments` reads now run through `mapWithConcurrency(..., 5, ...)` — bounded, input order preserved, per-ticket `try/catch` still isolating failures. Tested by peak-concurrency instrumentation: serialized would peak at 1, unbounded at N; the assertion requires `1 < peak <= 5`.

**Measured caveat, recorded in the code:** this buys real wall-clock only on the Linear adapter (`await fetch`). `github.mjs` shells out via `spawnSync`, which blocks the event loop, so its reads serialize regardless — 13.0s concurrent vs 12.5s serial for five held tickets on the factory repo, and 43s to 40s for the whole section. The fan-out here is correct and bounded; making it *pay* on GitHub is an adapter-level fix, filed as #1113. The comment says so, and says not to "fix" it by raising the limit.

## Also fixed (found while verifying)

A held comment read that **failed** rendered identically to a ticket with no comments — the section-level "unavailable vs empty" mistake repeated per row, and rate limits hit these reads one at a time so it is the common case. Now `comment unreadable — <reason>` vs `<title> (no comments)`.

## Verification

```
bun test --timeout 20000 orchestrator/ask.test.mjs && ./bin/factory ask --json | head -40 && ./bin/factory ask --section queue,held
```

11 pass / 0 fail / 112 expect() calls. The live run happened to hit the GitHub GraphQL secondary rate limit, which demonstrated the partial-failure path end to end: `legalease` returned 21 queued and 4 held, `factory` failed, the renderer printed `partial — factory: gh api graphql failed (status 1)`, and the command still exited 0.

Every new behaviour was observed failing against a deliberately broken build first (5 mutations: old `claimedAt` precedence + `heartbeatOf` without the `updatedAt` branch; `questionSource` and the renderer qualification removed; `HELD_COMMENT_CONCURRENCY = 1`; `unknown` rendered as `never`; the failed-read fallback collapsed into the title). Each killed the intended test and only that test.

## Filed to triage, not fixed here

- #1110 — a partially-failed section reads as authoritative to a JSON consumer (`error` is null when only some repos failed). Reproduced live on this very run.
- #1111 — `lib/spend.mjs` cannot distinguish "nothing spent" from "log dir unreadable"; `budgetExhausted` fails open on it.
- #1112 — `QUEUE — dispatchable now` overstates; it is the tracker predicate, not what `next` would start.
- #1113 — the GitHub adapter's `spawnSync` transport serializes every read, defeating caller concurrency.
- #1114 — read-only `runtime.db` open fails on an uncleanly-shut-down WAL database.

### DIFF
--- a/orchestrator/ask.mjs
+++ b/orchestrator/ask.mjs
@@ -633,6 +633,20 @@ const count = (section) =>
  * is only `updatedAt` says so, because "silent for 40m" and "last touched 40m
  * ago" are different claims about whether an agent is alive.
  */
+/**
+ * A comment read that FAILED must not render like a ticket that simply has no
+ * comments — falling back to the title in both cases is the section-level
+ * "unavailable vs empty" mistake repeated per row. Rate limits hit these reads
+ * one ticket at a time, so this is the common case, not the exotic one.
+ */
+const heldRowText = (r, repoW, idW) => {
+  const head = `  ${pad(r.repo, repoW)}  ${pad(r.identifier, idW)}  ${pad(r.holds.join(","), 22)}  `;
+  if (r.questionError)
+    return `${head}comment unreadable — ${oneLine(r.questionError, 60)}`;
+  if (r.question) return `${head}last comment: ${oneLine(r.question, 60)}`;
+  return `${head}${oneLine(r.title, 70)} (no comments)`;
+};
+
 const heartbeatText = (row) => {
   if (row.heartbeatSource === "comment")
     return `heartbeat ${humanAge(row.heartbeatAgeMs)} ago`;
@@ -687,12 +701,8 @@ export function formatAsk(doc) {
     // "what is waiting on me" into a lie.
     if (doc.held.rows?.length)
       lines.push("  (newest comment shown — may be the reply, not the ask)");
-    renderRows(
-      lines,
-      doc.held,
-      "nothing held",
-      (r) =>
-        `  ${pad(r.repo, repoW)}  ${pad(r.identifier, idW)}  ${pad(r.holds.join(","), 22)}  ${r.question ? `last comment: ${oneLine(r.question, 60)}` : oneLine(r.title, 70)}`,
+    renderRows(lines, doc.held, "nothing held", (r) =>
+      heldRowText(r, repoW, idW),
     );
   }
 
--- a/orchestrator/ask.mjs
+++ b/orchestrator/ask.mjs
@@ -73,6 +73,24 @@ export const WRITE_OPS = Object.freeze([
 
 const DEFAULT_WINDOW_HOURS = 24;
 
+/**
+ * How many `listComments` reads for held tickets are in flight at once.
+ *
+ * Small on purpose: this is a read surface that may be polled, and the point
+ * is to stop paying N round trips in series, not to burst a tracker's rate
+ * limiter — which, when it trips, is exactly the failure this command is
+ * supposed to survive.
+ */
+const HELD_COMMENT_CONCURRENCY = 5;
+
+/** `Promise.all` over fixed-size chunks: bounded fan-out, input order kept. */
+async function mapWithConcurrency(items, limit, fn) {
+  const out = [];
+  for (let i = 0; i < items.length; i += limit)
+    out.push(...(await Promise.all(items.slice(i, i + limit).map(fn))));
+  return out;
+}
+
 const reasonOf = (error) =>
   error?.message ? String(error.message) : String(error);
 
@@ -98,6 +116,26 @@ const msSince = (iso, now) => {
   const at = Date.parse(iso);
   return Number.isFinite(at) ? Math.max(0, now - at) : null;
 };
+
+/**
+ * When this ticket last showed a sign of life, and how confident that is.
+ *
+ * `lastCommentAt` is a real heartbeat — an agent posting its phase change.
+ * `updatedAt` is a proxy: it advances on the claim mutation, on label swaps
+ * and on comments, so it is an upper bound on silence rather than a heartbeat.
+ * The GitHub adapter pins `lastCommentAt: null` (github.mjs), which is the
+ * whole reason the distinction has to be carried rather than collapsed.
+ *
+ * Reporting a missing heartbeat as `never` would be the same mistake this
+ * command exists to prevent, one level down: "we did not read it" rendered as
+ * "it did not happen". Hence the explicit `unknown` source.
+ */
+export function heartbeatOf(ticket) {
+  if (ticket?.lastCommentAt)
+    return { at: ticket.lastCommentAt, source: "comment" };
+  if (ticket?.updatedAt) return { at: ticket.updatedAt, source: "updatedAt" };
+  return { at: null, source: "unknown" };
+}
 
 /** Compact duration for the human view: 5s / 12m / 3h / 2d. */
 export function humanAge(ms) {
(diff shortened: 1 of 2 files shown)


hdkiller:
## CI status: red, and it is base-caused — not this diff

Do not read the red X on this PR as a verdict on the change.

**Evidence:**

| | this PR (`25dfc56a`) | `develop` at this branch point (`1d7e0fc8`, run 33247393558) |
|---|---|---|
| Fast unit tests | failure | **failure** |
| Verify | failure | **failure** |
| Timing-bound tests | success | success |
| Shadow runner fleet available | success | success |

The same two jobs fail on the unmodified base commit this branch is cut from.

**The failing test is `event-runtime/lib/adapters/pi.test.mjs`** — `execute conformance (WM-440) > executes stub binary in workspaceDir, passes CURSOR_API_KEY, prompt on argv, captures transcript + trace`. This diff touches exactly two files, `orchestrator/ask.mjs` and `orchestrator/ask.test.mjs`, neither of which that test loads. Run locally on this branch, that file is **51 pass / 0 fail**.

**Two separate failures were involved, in sequence:**

1. The first run failed for a different reason — zero self-hosted runners were registered, so `Shadow runner fleet available` was cancelled after a 3-minute wait, which skipped `Fast unit tests`, `Full verification` and `Timing-bound tests` and failed the `Verify` aggregator. The fleet has since recovered (`watt-mind-runner-smoke-3/4/5` back online), and a re-run cleared it.
2. The re-run surfaced the real, pre-existing base failure above.

Per the coordinator, `develop` went red after five PRs were merged in ~2 minutes, including one merged while its own `Verify` and `Fast unit tests` were already failing. That is under separate diagnosis with a revert decision pending, so **this branch has deliberately not been rebased onto `develop`** and the failure has not been chased — it is outside this ticket's Owned Paths (`orchestrator/ask.mjs`, `orchestrator/ask.test.mjs`, `bin/factory`).

**This PR's own verification passes:** `bun test --timeout 20000 orchestrator/ask.test.mjs` — 11 pass / 0 fail / 112 expect() calls.

Recommend re-running CI on this PR once `develop` is green again, before any merge decision.

PR #1225, opened 2026-08-29
fix(connectors): block prototype-polluting secret paths

### LINKED ISSUE #1205: sec(connectors): private setAt/deleteAt in connectors.mjs lack own-property checks; moveHeuristic derives paths from config values
## Problem

`event-runtime/lib/connectors.mjs` (`splitConfigSecrets`, lines ~303-357 on `origin/develop`) moves secret-shaped values out of connector config into a secrets object using private `setAt`/`deleteAt` helpers. Those helpers walk `node[key]` with only an `isPlainObject` check and no own-property check, and `moveHeuristic` derives the key path from config **values** (any key matching `SECRET_KEY_HEURISTIC = /nsec|token|secret|key|password/i`). A connector config such as `{ "__proto__": { "token": "x" } }` or `{ "constructor": { "prototype": { "apiKey": "x" } } }` therefore walks the traversal into `Object.prototype` and can pollute it (or read an inherited value as if it were config).

Not reachable through schema-derived paths, and PR #1200 (open, `extensions.mjs`) hardens the sibling helpers there; the connectors copies need the same treatment independently of #1200 landing. Found in review of PR #1200.

## Fix

In `event-runtime/lib/connectors.mjs` only:

- Reject the forbidden segments `__proto__`, `constructor`, and `prototype` at every depth in `setAt`, `deleteAt`, and `moveHeuristic` (skip the key with no write; do not throw on user config).
- `setAt`/`deleteAt` descend only into own properties (`Object.hasOwn(node, key)`), and `setAt` creates intermediate nodes as `Object.create(null)` or `{}` assigned via own-property write only.
- `moveHeuristic` iterates own enumerable keys only (`Object.entries` already does; keep it) and never recurses into a forbidden segment.

Do not change the exported signature of `splitConfigSecrets`; do not touch `extensions.mjs` (owned by #1200).

## Acceptance Criteria

- [ ] `splitConfigSecrets({ __proto__: { token: "x" } })` and `splitConfigSecrets(JSON.parse('{"__proto__":{"token":"x"}}'))` leave `Object.prototype.token` undefined and return `secrets` without a `__proto__` key.
- [ ] `splitConfigSecrets(JSON.parse('{"constructor":{"prototype":{"apiKey":"x"}}}'))` leaves `Object.prototype.apiKey` undefined.
- [ ] Explicit `secretFields` whose `path` contains a forbidden segment are ignored without throwing and without a write.
- [ ] `setAt`/`deleteAt` never descend through inherited properties (a test with an object whose prototype carries a plain-object `nested` field shows `secrets.nested` is created fresh, not written into the prototype's object).
- [ ] Existing `splitConfigSecrets` behavior (format: secret fields and heuristic keys move; non-secret values stay) is unchanged and covered by the existing tests.
- [ ] New tests live in `event-runtime/lib/connectors.test.mjs`; each asserts on `Object.prototype` and cleans up any pollution in `afterEach` so a regression cannot poison later tests.

## Owned Paths

- `event-runtime/lib/connectors.mjs`
- `event-runtime/lib/connectors.test.mjs`

## Verification Command

```bash
bun test event-runtime/lib/connectors.test.mjs --timeout 20000 && bun run check
```

hdkiller:
Fixes watt-mind/factory#1205

Harden connector secret-path traversal against prototype pollution without changing the exported splitConfigSecrets API.

## Validation

| Verification Command | `bun test event-runtime/lib/connectors.test.mjs --timeout 20000 && bun run check` | pass    | 17 tests, 0 failures; generated files match |
| Repo verify          | `bun test event-runtime/lib --timeout 20000 && bun build/emit.mjs --check` | pass    | event-runtime/lib suite and generated-files check completed |
| UX critique          | factory-ux-critic                | not run | backend/security hardening; no user-facing surface |

UX critique: skipped — backend/security hardening; no user-facing surface

### DIFF
--- a/event-runtime/lib/connectors.mjs
+++ b/event-runtime/lib/connectors.mjs
@@ -57,6 +57,11 @@ export const CONNECTOR_ID_PATTERN =
 export const CONNECTOR_START_TIMEOUT_MS = 10_000;
 
 const SECRET_KEY_HEURISTIC = /nsec|token|secret|key|password/i;
+const FORBIDDEN_CONFIG_PATH_SEGMENTS = new Set([
+  "__proto__",
+  "constructor",
+  "prototype",
+]);
 
 /** Typed error for a module that fails the connector contract. */
 export class ConnectorError extends Error {
@@ -304,6 +309,7 @@ export function splitConfigSecrets(values, secretFields = []) {
   const config = cloneJson(values) ?? {};
   const secrets = {};
   const move = (keyPath) => {
+    if (keyPath.some((key) => FORBIDDEN_CONFIG_PATH_SEGMENTS.has(key))) return;
     let node = values;
     for (const key of keyPath) {
       if (
@@ -326,6 +332,9 @@ function moveHeuristic(from, secrets, path) {
   if (!from || typeof from !== "object" || Array.isArray(from)) return;
   for (const [key, inner] of Object.entries(from)) {
     const next = [...path, key];
+    if (next.some((segment) => FORBIDDEN_CONFIG_PATH_SEGMENTS.has(segment))) {
+      continue;
+    }
     if (SECRET_KEY_HEURISTIC.test(key) && !isPlainObject(inner)) {
       setAt(secrets, next, inner);
       delete from[key];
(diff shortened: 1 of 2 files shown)


hdkiller:
Closing because this ticket changes connector secret handling and requires human authorization under the dispatch policy's secrets escalation. No merge performed.

Issue #1212, opened 2026-08-29
refactor(worker): derive releaseStalledWorkerLease retry cause via retryDecision instead of hardcoded retry:environment


hdkiller:
## Problem & Context

`releaseStalledWorkerLease` (`event-runtime/lib/worker.mjs` ~L3519-3615) is the operator recovery path for a worker that stopped heartbeating while owning a run. Its retry branch hardcodes two things that the rest of the worker derives from the failure taxonomy:

- it decides whether to requeue with `run.attempts < spec.maxAttempts`, whereas the reaper (`worker.mjs` ~L3682) and the attempt finalizers (~L2147, ~L3120) ask `retryDecision(db, runId, spec, "lease_expired")`, which honours `maxEnvironmentRetries` for environment-classed failures;
- it emits the QUEUED transition with the literal `reason: "retry:environment"` instead of `retry:${decision.cause}`.

If `ENVIRONMENT_FAILURES` or `classifyFailureCause` ever reclassifies `lease_expired`, or `maxEnvironmentRetries` differs from `maxAttempts`, the operator path silently drifts from the reaper. Found reviewing PR #1196 (#947).

Note: `worker.mjs` ~L1363 (`requeue?.reason !== "retry:environment"`) consumes the literal, so the emitted string for the environment cause must remain exactly `retry:environment`.

## Acceptance Criteria

- [ ] `releaseStalledWorkerLease` calls `retryDecision(db, heldRunId, spec, "lease_expired")` after `finishAttempt(...)` (so the count includes the current failure) and requeues only when `decision.retry` is true; the `run.attempts < spec.maxAttempts` guard is removed.
- [ ] The QUEUED transition reason is `` `retry:${decision.cause}` `` (still `retry:environment` for `lease_expired`); the FAILED reason still comes from `typedFailureReason("lease_expired")`.
- [ ] When `decision.retry` is false the existing exhaustion branch (RUNNING -> FAILED dead-letter) runs unchanged.
- [ ] New test in `event-runtime/lib/worker.test.mjs`: a spec with `maxAttempts: 5, maxEnvironmentRetries: 0` and one prior `lease_expired` failure is dead-lettered by `releaseStalledWorkerLease` instead of requeued (proves the decision, not the literal, drives the branch).
- [ ] Existing `releaseStalledWorkerLease` tests (~L2441-2500) and `event-runtime/lib/reaper.test.mjs` stay green; the docblock above the function no longer says "fixed `retry:environment` reason".

[...]

PR #1207, opened 2026-08-29
fix(security): bind pack prompts to verified bytes

hdkiller:
Fixes watt-mind/factory#965

Review the pack containment boundary and verified prompt snapshot handoff; security authorisation `operator:preapproved-2026-08-29` decided `2026-08-29T18:05:43.689Z` matched the current description.

## Validation

| Check                | Command                                                                                           | Result  | Notes                  |
| -------------------- | ------------------------------------------------------------------------------------------------- | ------- | ---------------------- |
| Verification Command | `bun test --timeout 20000 event-runtime/lib/registry.test.mjs event-runtime/lib/api-registry.test.mjs` | pass    | 72 tests, 0 failures   |
| Repo verify          | `bun test event-runtime/lib --timeout 20000 && bun build/emit.mjs --check`                         | pass    | 1709 pass, emit clean  |
| UX critique          | factory-ux-critic                                                                                 | not run | no user-facing surface |

UX critique: skipped — no user-facing surface
- Post-review (58283319): merged origin/develop (worker.mjs both sides kept); `createFsPackLoader` now raises `RegistryError` for a missing pack root instead of raw ENOENT; registry/adapters/worker tests (460 pass) and `bun run check` green.

PR #1227, opened 2026-08-29
fix(worker): stop permanent artifact retries

### LINKED ISSUE #1187: environment-classed run failures retry past maxAttempts (permanent 'artifact not in store' loops forever)
## Problem & Context

A run that fails with `failure:environment:adapter_error: artifact <sha> is not in the store` is re-queued **past its maxAttempts** and retries indefinitely. Observed a `dispatch@1` run at **`attempts 4/1`** (maxAttempts=1) still cycling RUNNING → FAILED → requeue every ~30s; it does not self-terminate.

Two things are wrong:

1. **Environment-classed failures have their own budget, and it does not hold.** `retryDecision()` in `event-runtime/lib/worker.mjs` (~line 795) retries `environment` causes while `failureCount(...) <= maxEnvironmentRetries(spec)` (`DEFAULT_MAX_ENVIRONMENT_RETRIES = 3`, worker.mjs:706), independent of `spec.maxAttempts`. Either the count is not reached because attempts are being requeued by another path (reaper / planner `includeFailed` requery at planner.mjs:1575-1600), or the ceiling is being re-evaluated against a reset count. The observed run exceeded both `maxAttempts` (1) and `maxEnvironmentRetries + 1` (4), so at least one requeue path is unbounded.
2. **A missing *input* artifact is permanent, not transient.** `workspace.mjs:218` / `artifacts.mjs:273` throw `artifact <sha> is not in the store`; the worker maps that to the generic `adapter_error` reason code (worker.mjs:3400-3411), which `classifyFailureCause` treats as `environment` → retryable. Retrying can never succeed.

Repro context: 2026-08-29 after a serve restart, on runs whose input artifact was lost to a store-consistency incident (`62fcdc79…`). Runs `run_0438cc6d…`, `run_5a255546…`. Poisoned runs burn worker leases forever and flood serve.log. `cli cancel <run-id>` cannot stop them (`illegal transition FAILED → CANCELLED`).

## Acceptance Criteria

- [ ] A missing input artifact (`... is not in the store` raised while materializing inputs) gets its own reason code (e.g. `input_artifact_missing`) that `classifyFailureCause` returns `fatal` for: the run goes terminal FAILED on the first occurrence and is never requeued.
- [ ] Environment-classed failures are bounded on **every** requeue path: after `maxEnvironmentRetries + 1` finished attempts with an environment cause, the run stays FAILED. A regression test drives a run with `maxAttempts: 1` through repeated `adapter_error` failures and asserts the attempt count never exceeds the ceiling.
- [ ] The journal reason on the terminal attempt names why retries stopped (permanent cause vs. environment budget exhausted).
- [ ] `retryRun()` semantics for operators are unchanged apart from the new fatal code (which `--force` may still override, like other fatal codes).
- [ ] Out of scope, file separately if not already: an operator escape hatch (`cancel --force` from FAILED).

## Source File Pointers

- `event-runtime/lib/worker.mjs` — `ENVIRONMENT_FAILURES` / `FATAL_FAILURES` / `classifyFailureCause` (~740-780), `retryDecision` (~795), reason-code mapping at ~3400-3411, `retryRun` (~3806)
- `event-runtime/lib/worker.test.mjs` — existing failure-taxonomy and retry coverage
- `event-runtime/lib/workspace.mjs:218`, `event-runtime/lib/artifacts.mjs:273` — where the error originates (read only; not owned)
- `event-runtime/lib/planner.mjs:1575-1600` — `includeFailed` requery predicate (read only; not owned)

## Owned Paths

- event-runtime/lib/worker.mjs
- event-runtime/lib/worker.test.mjs

## Verification Command

```
bun test --timeout 20000 --max-concurrency=4 event-runtime/lib/worker.test.mjs
```

hdkiller:
Fixes watt-mind/factory#1187

Missing input artifacts now fail terminally, and exhausted environment retries record their terminal cause.

## Validation

| Verification Command | `bun test --timeout 20000 --max-concurrency=4 event-runtime/lib/worker.test.mjs` | pass | 140 tests, 0 failures |
| Repo verify          | `bun test event-runtime/lib --timeout 20000 && bun build/emit.mjs --check` | pass | 1724 tests, emit clean |
| UX critique          | factory-ux-critic                | not run | no user-facing surface |

UX critique: skipped

### Post-review

- Merged `origin/develop` (worker.mjs rewritten by #1207 pack-prompt/def-hash gate and #1215 workspace confinement / `guestEnvironment` / `isSandboxGuarded`); conflict was only the `createWorkspace` call site — kept both: develop's earlier `adapterKey` binding plus this branch's narrow try/catch classifying `artifact <sha256> is not in the store` as `input_artifact_missing` (fatal). Prettier applied.
- AC#2 ("bounded on every requeue path") is only partially met: `releaseStalledWorkerLease` still requeues on bare `attempts < maxAttempts` without consulting `retryDecision`. Tracked by #1212.

### DIFF
--- a/event-runtime/lib/worker.mjs
+++ b/event-runtime/lib/worker.mjs
@@ -802,22 +802,42 @@ function failureCount(db, runId, cause) {
 }
 
 /** Called after the current attempt is finalized, so counts include this failure. */
-function retryDecision(db, runId, spec, reasonCode) {
+function retryDecision(
+  db,
+  runId,
+  spec,
+  reasonCode,
+  { includeCurrentFailure = false } = {},
+) {
   const cause = classifyFailureCause(reasonCode);
   if (cause === "environment") {
+    const failures =
+      failureCount(db, runId, cause) + (includeCurrentFailure ? 1 : 0);
     return {
       cause,
-      retry: failureCount(db, runId, cause) <= maxEnvironmentRetries(spec),
+      retry: failures <= maxEnvironmentRetries(spec),
     };
   }
   if (cause === "agent_error") {
-    return { cause, retry: failureCount(db, runId, cause) < spec.maxAttempts };
+    const failures =
+      failureCount(db, runId, cause) + (includeCurrentFailure ? 1 : 0);
+    return { cause, retry: failures < spec.maxAttempts };
   }
   return { cause, retry: false };
 }
 
 function typedFailureReason(reasonCode, detail = reasonCode) {
   return `failure:${classifyFailureCause(reasonCode)}:${detail}`;
+}
+
+function terminalFailureReason(decision, reasonCode, detail = reasonCode) {
+  if (decision.cause === "environment" && !decision.retry) {
+    return typedFailureReason(
+      reasonCode,
+      `${detail}; environment_retry_budget_exhausted`,
+    );
+  }
+  return typedFailureReason(reasonCode, detail);
 }
 
 function resolveNow(now) {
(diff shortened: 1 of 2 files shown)


hdkiller:
factory-merge-fix round=1 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=39e2c37b9880ac400f04b8a46e5839c25550bdb2 new=484b69fce3a29e2ac725dc82ba698c8a72a9b263

PR #1348, opened 2026-08-29
fix(worker): report terminal path failures (#1330)

### LINKED ISSUE #1330: fix(worker): executeClaimed swallows failTerminal/finishAttempt/unclaim failures on the terminal path — log and trace them
## Context
`executeClaimed` (`event-runtime/lib/worker.mjs:2371`) has three terminal-path failures that are swallowed with no log line, no trace record and no return signal:
- `worker.mjs:3396-3407`: on cancel/force-fail, `unclaimTicketFn(...)` is wrapped in `catch { /* intentionally ignored */ }` with `log: null` — if the tracker unclaim fails the ticket stays claimed with nobody told.
- `worker.mjs:3425-3437`: `finishAttempt(db, runId, attempt, "CANCELLED", …)` is wrapped in `catch { // ignore }` — the attempt row keeps `finished_at = NULL` and the run reports `{ ok: true }` anyway.
- `worker.mjs:4110-4115`: after an adapter error, `res = failTerminal("FAILED", journalReason, reasonCode)` is wrapped in `catch { // if failTerminal could not transition, continue }` — the run can be left in RUNNING/VERIFYING with the workspace already cleaned (`cleanupWorkspace` follows on `:4116`), and `res` is undefined so the fenced check is skipped.
The rest of this file routes best-effort failures through `recordTrace`/`appendTrace` (e.g. `worker.mjs:3288`) or `console.error`; these three do neither, so a lease-reaper-only recovery is the first sign anything went wrong.

## Acceptance Criteria
- [ ] Each of the three catches emits a `console.error` line with `runId`, `attempt` and `err.message` and, where an attempt row exists, an `attempt_trace` record of kind `terminal_error` (or the nearest existing kind) carrying the same fields.
- [ ] `worker.test.mjs` covers the `failTerminal` case: an adapter failure whose `failTerminal` throws yields a logged line and a trace record, and the run does not silently return `{ ok: true }`.
- [ ] `worker.test.mjs` covers the cancel case: a throwing `finishAttempt` produces a logged line and the returned object flags the failure (e.g. `finishError`) instead of `{ ok: true }`.

## Owned Paths
- event-runtime/lib/worker.mjs
- event-runtime/lib/worker.test.mjs

## Verification Command
`bun test event-runtime/lib/worker.test.mjs --timeout 20000`

hdkiller:
Fixes watt-mind/factory#1330

Review terminal-path error observability and retry signals before merging.

## Validation

| Check                | Command                                                                                             | Result  | Notes                              |
| -------------------- | --------------------------------------------------------------------------------------------------- | ------- | ---------------------------------- |
| Verification Command | `bun test event-runtime/lib/worker.test.mjs --timeout 20000`                                      | pass    | 163 tests, 0 failures              |
| Repo verify          | `bun test event-runtime/lib --timeout 20000 && bun build/emit.mjs --check && bun run format:check` | pass    | 1890 pass, 10 skipped, generated clean |
| UX critique          | factory-ux-critic                                                                                   | skipped | no user-facing backend surface     |

UX critique: skipped

## Post-review

- Review verdict: MERGE with one nit, addressed in a140e83c: `traceRecorder(db, {runId, attempt})` is now constructed inside a try/catch at `executeClaimed` entry; a prepare failure logs and falls back to a no-op recorder with the same call surface (`recorder(kind, payload)`, `recorder.stats()`), so it can no longer throw out of `executeClaimed`. Covered by a new worker test that drops `attempt_trace` before `runOnce` and asserts the run still COMPLETES.
- Merged `origin/develop` (no conflicts). Verified: `bun test event-runtime/lib` (1900 pass), `bun build/emit.mjs --check`, `format:check`, `bun run check` all green.

### DIFF
--- a/event-runtime/lib/worker.mjs
+++ b/event-runtime/lib/worker.mjs
@@ -2703,6 +2703,26 @@ export async function executeClaimed(
   // auditor needs, so it must not be attached to the success path alone.
   // Null until admission runs, and null for every confined run.
   let filesystemConfinementReceipt = null;
+  const recorder = traceRecorder(db, { runId, attempt });
+  const recordTerminalError = (operation, err) => {
+    const message = err?.message ?? String(err);
+    console.error(
+      `[worker] terminal ${operation} failed for run ${runId} attempt ${attempt}: ${message}`,
+    );
+    // `lifecycle` is the nearest supported trace kind to a terminal error.
+    try {
+      recorder("lifecycle", {
+        terminalError: true,
+        operation,
+        runId,
+        attempt,
+        message,
+      });
+    } catch {
+      // A terminal error must still be reported when trace storage is down.
+    }
+    return message;
+  };
 
   const refuseTerminal = (
     reasonCode,
@@ -3363,7 +3383,6 @@ export async function executeClaimed(
     // Real wall-clock per event — NOT the claim-time `now`. Trace timestamps
     // are the one place frozen time defeats the feature: "what is the agent
     // doing right now" needs to say when each step actually happened.
-    const recorder = traceRecorder(db, { runId, attempt });
     const onTrace = (kind, payload) => {
       try {
         recorder(kind, payload);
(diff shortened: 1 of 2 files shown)


hdkiller:
factory-merge-fix round=1 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=5ced445acad90df176823c82cc2a63d45e9e2d7c new=353fa231806fecce8d2a219af475eb27a6f3253b

PR #1492, opened 2026-08-30
fix(backend): clamp GitHub control-plane bodies

### LINKED ISSUE #1482: fix(backend): github control plane sends comment/issue bodies over the 65536-char limit — a long report 422s and is lost
## Context
- GitHub rejects issue and comment bodies longer than 65 536 characters with `422 Validation Failed: body is too long (maximum is 65536 characters)`.
- `lib/control-plane/github.mjs:1738-1745` — `comment()` POSTs `stampRun(body)` with no length guard; `:1817-1822` — `file()` POSTs `body` unguarded; `:1842-1853` — `appendDetail()` PATCHes `issue.body + detail` unguarded, so every append grows the body until the ticket becomes un-editable; `:1422-1431` — the ready-pin comment path likewise.
- `lib/control-plane/labels.mjs` `stampRun()` appends `\n\nrun:<id>` *after* the body, so a naive truncation would drop the attribution trailer.
- A long handoff/verify report (test logs are routinely >64 KiB) therefore fails the whole `comment` call with a `ControlPlaneError` and the run's report is lost instead of landing truncated.

## Acceptance Criteria
- [ ] A shared helper (e.g. `clampGithubBody(body, {max: 65536})` in `labels.mjs`) truncates over-long markdown with a visible `… [truncated N chars]` marker while keeping the `run:<id>` trailer and not splitting inside a fenced code block.
- [ ] `comment()`, `file()`, `appendDetail()` and the ready-pin comment path use the helper; `appendDetail()` reports `{appended:false, reason:"body_full"}` when the existing body leaves no room, rather than throwing after the PATCH.
- [ ] `lib/control-plane/github.test.mjs` covers each path with a 70 000-char body against the fake `api` seam and asserts the sent body length ≤ 65 536 and the trailer present.
- [ ] `bun test lib/control-plane/github.test.mjs --timeout 20000` passes.

## Owned Paths
- lib/control-plane/labels.mjs

## Verification Command
`bun test lib/control-plane/github.test.mjs --timeout 20000`

hdkiller:
Fixes watt-mind/factory#1482

Clamp GitHub control-plane report bodies before a 422 can discard the report.

## Validation

| Check                | Command                                                           | Result  | Notes                                |
| -------------------- | ----------------------------------------------------------------- | ------- | ------------------------------------ |
| Verification Command | `bun test lib/control-plane/github.test.mjs --timeout 20000`     | pass    | 105 tests, 0 failures                |
| Repo verify          | `bun test event-runtime/lib --timeout 20000 && bun build/emit.mjs --check && bun run format:check` | pass    | 1993 pass, 10 skipped, emit/format clean |
| UX critique          | factory-ux-critic                                                 | skipped | backend control-plane; no user-facing surface |

UX critique: skipped

## Post-review

- `clampGithubBody` now cuts inside a fenced block and closes the fence with the same delimiter before the marker; the fence is only dropped when none of its content fits. A report that is mostly one long log keeps the log up to the budget (was: 72,061 chars → 75).
- Budget is measured in UTF-8 bytes (`Buffer.byteLength`) and the cut lands on a code-point boundary, never splitting a surrogate pair. `appendDetail` computes its remaining budget in bytes too.
- A `max` too small for the bare marker returns `""` instead of a mangled partial marker; `appendDetail` already reports `body_full` on that.
- The `run:` trailer is carved out only when `FACTORY_COMMENT_ATTRIBUTION=1` and the body ends with this run's exact stamp.
- Tests: fenced log survives truncated, closed fence + prose, fence dropped when nothing fits, byte budget / surrogate safety, empty result on tiny max, trailer anchoring.

### DIFF
--- a/lib/control-plane/github.mjs
+++ b/lib/control-plane/github.mjs
@@ -1853,8 +1853,12 @@ export function githubControlPlane(options = {}) {
       const rawDetail = String(markdown).trim();
       const suffix = description.slice(current.length);
       const separator = suffix.slice(0, -rawDetail.length - 1);
+      // Byte budget, matching clampGithubBody (never smaller than chars).
       const available =
-        GITHUB_BODY_MAX_LENGTH - current.length - separator.length - 1;
+        GITHUB_BODY_MAX_LENGTH -
+        Buffer.byteLength(current, "utf8") -
+        Buffer.byteLength(separator, "utf8") -
+        1;
       if (available <= 0) return { appended: false, reason: "body_full" };
       const detail = clampGithubBody(rawDetail, { max: available });
       if (!detail) return { appended: false, reason: "body_full" };
--- a/lib/control-plane/github.mjs
+++ b/lib/control-plane/github.mjs
@@ -24,7 +24,9 @@ import {
   AGENT_READY_LABEL,
   appendIssueDetail,
   BLOCKED_LABEL,
+  clampGithubBody,
   claimLabels,
+  GITHUB_BODY_MAX_LENGTH,
   IN_PROGRESS_LABEL,
   stampRun,
   validateLabels,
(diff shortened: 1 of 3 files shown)


hdkiller:
factory-merge-fix round=1 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=0a071b40fb9f69b5100ddad446db4f815aea6eb1 new=9dfc05cffe6d89fa676e0decdb82ac3e11131824

PR #1512, opened 2026-08-30
fix(control-plane): paginate GitHub labels

### LINKED ISSUE #1484: fix(backend): github repoLabels reads one page — repos with >100 labels get false "label does not exist" refusals
## Context
- `lib/control-plane/github.mjs:879-884` — `repoLabels(repo)` does a single `GET repos/:repo/labels?per_page=100` and returns that page only.
- `lib/control-plane/github.mjs:886-898` — `requireLabelsExist()` compares the requested labels against that first page and throws `label(s) do not exist in this workspace: … fix: factory init --control-plane github` for any label not on it.
- A repo with more than 100 labels (the factory protocol set is ~40 on top of GitHub defaults plus `tier:*`, `area:*`, `priority:*` families) therefore gets a false "label does not exist" refusal on `file()` (`:1816`) and on `claim`/`setLabels` label writes, and the suggested `factory init` fix reports every label as `existed`.
- The same file already has a paginating helper `listAllPages(pathName, query)` at `:964-982` (PAGE_SIZE/MAX_PAGES, safety ceiling warning) used for issues; `repoLabels` predates it.

## Acceptance Criteria
- [ ] `repoLabels()` uses `listAllPages("repos/:repo/labels")` so every page is consulted; the result is cached per repo for the lifetime of the control-plane instance (labels change rarely; `file()` and `claim()` call it repeatedly).
- [ ] `provisionGithubRepo()` (`:1946-1990`) uses the same paginated read for its `existing` set so init does not re-create labels on page 2+ (which currently surfaces as swallowed 422s).
- [ ] `lib/control-plane/github.test.mjs` adds a fake-`api` case with 150 labels where the requested label sits on page 2 and asserts no `ControlPlaneError`, plus a case asserting one label fetch across two `file()` calls.
- [ ] `bun test lib/control-plane/github.test.mjs --timeout 20000` passes.

## Owned Paths

## Verification Command
`bun test lib/control-plane/github.test.mjs --timeout 20000`

hdkiller:
Fixes watt-mind/factory#1484

Paginate and cache GitHub label reads so page-two labels can be used by ticket actions.

## Validation

| Check                | Command                                              | Result  | Notes                                  |
| -------------------- | ---------------------------------------------------- | ------- | -------------------------------------- |
| Verification Command | `bun test lib/control-plane/github.test.mjs --timeout 20000` | pass    | 110 tests, 0 failures                  |
| Repo verify          | `bun test event-runtime/lib --timeout 20000 && bun build/emit.mjs --check && bun run format:check` | pass    | 1996 pass, 10 skipped, emit and format clean |
| UX critique          | factory-ux-critic                                    | not run | backend-only control-plane change      |

UX critique: skipped — no user-facing surface

## Post-review

Reviewer verdict FIX, addressed in 3165f083:

1. `repoLabels(repo)`: a rejected `listAllPages` promise is evicted from `repoLabelsCache` (mirrors the `projectItems()` precedent), so a transient 403/timeout/rate-limit is no longer replayed to every later `file()`/`claim()`/`setLabels()` until restart. Test: label read throws once, second `file()` succeeds with a fresh read.
2. `requireLabelsExist`: on a computed miss, drop the memo and re-read once before refusing, so a label created by `factory init` takes effect without restarting serve. Test: miss triggers exactly one re-read; label added out of band is then found.
3. Restored the `listAllPages` docstring rationale (GitHub defaults to 30 records; callers must opt into pagination).

### DIFF
--- a/lib/control-plane/github.mjs
+++ b/lib/control-plane/github.mjs
@@ -120,7 +120,12 @@ const collaboratorAssociationCache = new Map();
 const PAGE_SIZE = 100;
 const MAX_PAGES = 50;
 
-/** Read a bounded GitHub REST collection in full. */
+/**
+ * Read a bounded GitHub REST collection in full. GitHub defaults to 30
+ * records, so callers that need the complete collection (comment history,
+ * repo labels) must opt into pagination rather than silently treating the
+ * first page as the entire collection.
+ */
 async function listAllPages(call, pathName, query = {}) {
   const items = [];
   for (let page = 1; page <= MAX_PAGES; page += 1) {
@@ -901,8 +906,16 @@ export function githubControlPlane(options = {}) {
   }
 
   async function repoLabels(repo) {
-    if (!repoLabelsCache.has(repo))
-      repoLabelsCache.set(repo, listAllPages(call, `repos/${repo}/labels`));
+    if (!repoLabelsCache.has(repo)) {
+      const promise = listAllPages(call, `repos/${repo}/labels`);
+      // A rejected read (403 mid token refresh, timeout, rate limit) must not
+      // be replayed to every later file()/claim()/setLabels() until restart:
+      // drop the memo so the next caller retries instead of inheriting it.
+      promise.catch(() => {
+        if (repoLabelsCache.get(repo) === promise) repoLabelsCache.delete(repo);
+      });
+      repoLabelsCache.set(repo, promise);
+    }
     return repoLabelsCache.get(repo);
   }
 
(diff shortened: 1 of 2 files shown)


hdkiller:
factory-merge-fix round=1 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=1147e7ff96e3a0d25223c2bb347af3eeeb1b1901 new=06a0f6d6a789d365e6fc0f64590f6118eb1585f6

PR #1557, opened 2026-08-30
perf(linear): avoid duplicate label catalog pagination

### LINKED ISSUE #1549: perf(linear): labelIdsFor paginates the whole label catalogue twice on every claim/transition
## Context

`lib/control-plane/linear.mjs:247-261` (`labelIdsFor`):

```js
let all = await allLabels();
let missing = add.filter((n) => !all.some((l) => l.name === n));
if (missing.length) all = await allLabels();
missing = add.filter((n) => !all.some((l) => l.name === n));
```

`allLabels()` (line 238-245) is an uncached, fully paginated `issueLabels(first:250)` walk of the whole workspace. The second call is byte-for-byte the same request against the same server state milliseconds later; it looks like a leftover from a cached-catalogue design that no longer exists. Consequences:

- Every `claim`, `transition` with labels, and `setLabels` already pays one full label-catalogue pagination; when a label genuinely does not exist (a misspelled `area:*`, `type:chore` slipping past `validateLabels`) the adapter pays it twice before failing — doubling the GraphQL cost of the failure path that the `gql` retry/backoff (`orchestrator/reaper.mjs:184`) is already stretching.
- `file()` (line 452) does its own `allLabels()` + missing check with a different error path, so the same rule is now implemented twice in one file.

The contract suite injects `gql`, so the number of `issueLabels` calls is observable and can be asserted.

## Acceptance Criteria

- [ ] `labelIdsFor` fetches the label catalogue exactly once per call; a missing label throws `label(s) do not exist in this workspace: …` after a single `issueLabels` query.
- [ ] `file()` reuses the same helper for the "labels exist" check so the message and behaviour are identical.
- [ ] A test in `lib/control-plane/linear.test.mjs` counts `issueLabels` queries through the injected `gql` for a claim/transition with a missing label and asserts exactly one; observed failing (two) before the fix.
- [ ] Existing linear contract tests pass unchanged.

## Owned Paths

- lib/control-plane/linear.mjs
- lib/control-plane/linear.test.mjs

## Verification Command

```
bun test --timeout 20000 lib/control-plane/linear.test.mjs
```

hdkiller:
Fixes watt-mind/factory#1549

Review the single-query label validation refactor and its missing-transition-label regression test.

## Validation

| Check                | Command                                                        | Result  | Notes                       |
| -------------------- | -------------------------------------------------------------- | ------- | --------------------------- |
| Verification Command | `bun test --timeout 20000 lib/control-plane/linear.test.mjs`  | pass    | 8 tests, 0 failures         |
| Repo verify          | `bun test event-runtime/lib --timeout 20000 && bun build/emit.mjs --check && bun run format:check` | pass    | 2032 pass, 0 failures       |
| UX critique          | factory-ux-critic                                              | not run | backend control-plane change |

UX critique: skipped

run:run_690a8c58-0755-4dfc-8e24-f8d5f9751db4

### DIFF
--- a/lib/control-plane/linear.mjs
+++ b/lib/control-plane/linear.mjs
@@ -248,10 +248,8 @@ export function linearControlPlane({
     const bad = validateLabels(add);
     if (bad.length)
       throw new ControlPlaneError("invalid label(s):\n  " + bad.join("\n  "));
-    let all = await allLabels();
-    let missing = add.filter((n) => !all.some((l) => l.name === n));
-    if (missing.length) all = await allLabels();
-    missing = add.filter((n) => !all.some((l) => l.name === n));
+    const all = await allLabels();
+    const missing = add.filter((n) => !all.some((l) => l.name === n));
     if (missing.length)
       throw new ControlPlaneError(
         `label(s) do not exist in this workspace: ${missing.join(", ")}`,
@@ -449,12 +447,6 @@ export function linearControlPlane({
       );
       if (!target)
         throw new ControlPlaneError(`team ${team} has no "${state}" state`);
-      const all = await allLabels();
-      const missing = labels.filter((n) => !all.some((l) => l.name === n));
-      if (missing.length)
-        throw new ControlPlaneError(
-          `label(s) do not exist in this workspace: ${missing.join(", ")}`,
-        );
       const d = await call(
         `mutation($in:IssueCreateInput!){ issueCreate(input:$in){ success issue{ identifier url } } }`,
         {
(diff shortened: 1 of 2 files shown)


hdkiller:
factory-merge-fix round=1 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=a5dc4a842ed9d76b39b556c76bef6540f53d1597 new=bbe63cae5e3ac29889ee1434aad59212d284a130

PR #1561, opened 2026-08-30
fix(bin): preserve live daemon log writers on rotation

### LINKED ISSUE #1547: bug(bin): worktree-daemons.sh rotates a live daemon log by rename — daemon keeps writing to .1 and the fresh log stays empty
## Context

`bin/worktree-daemons.sh:32-45` (`rotate_log_file`) rotates by rename:

```bash
mv -f "$logfile" "${logfile}.1"
touch "$logfile"
```

`rotate_daemon_logs` (line 48-61) calls it for `serve.log`, `worker.log`, `web.log`, and `supervise_daemons` (line 323-327) runs it every tick while the daemons are alive. The daemons were spawned by `spawn_daemon` (`bin/worktree-common.sh:453-478`) with stdout/stderr bound to the log's file descriptor, so after the rename they keep appending to the same inode — which is now `<log>.1`. The fresh `<log>` stays empty forever, `<log>.1` keeps growing without bound (it is never size-checked), and `factory tail` / `check_daemon_health` readers see an empty live log. On the next rotation `.1` is overwritten by another rename of an empty file, destroying the only copy of the output. `bin/worktree-common.sh:519-531` already documents exactly this hazard and implements the correct copy-truncate path (`rotate_run_log`, with `run_log_has_live_owner` and `keep` generations) — `live-stack.sh` uses it; `worktree-daemons.sh` has a parallel, older implementation that does not.

The existing test (`bin/worktree-daemons.test.mjs:211-236`) only checks the byte counts of the rotated files, so it cannot see the inode problem.

## Acceptance Criteria

- [ ] `rotate_log_file` / `rotate_daemon_logs` in `bin/worktree-daemons.sh` delegate to `rotate_run_log` / `rotate_run_logs` from `worktree-common.sh` (copy-truncate when a live pidfile owner exists, rename otherwise, `keep` generations), with the default 10 MiB threshold preserved.
- [ ] A test starts a background writer whose stdout is redirected to `serve.log`, writes a `serve.pid` with its pid, rotates, then has the writer emit one more line and asserts that line lands in `serve.log` (not in `serve.log.1`). Observed failing before the fix.
- [ ] The existing byte-count test and the `rotate-logs` CLI subcommand test still pass.
- [ ] `--help` text and the header comment for `rotate-logs` describe copy-truncate for live daemons.

## Owned Paths

- bin/worktree-daemons.sh
- bin/worktree-daemons.test.mjs

## Verification Command

```
bun test --timeout 60000 bin/worktree-daemons.test.mjs bin/worktree-common.test.mjs
```

hdkiller:
Fixes watt-mind/factory#1547

Route daemon log rotation through the live-owner-aware helper so active writers keep appending to the current log.

## Validation

| Check                | Command                                                                          | Result  | Notes                              |
| -------------------- | -------------------------------------------------------------------------------- | ------- | ---------------------------------- |
| Verification Command | `bun test --timeout 60000 bin/worktree-daemons.test.mjs bin/worktree-common.test.mjs` | pass    | 55 tests, 0 failures               |
| Repo verify          | `bun test event-runtime/lib --timeout 20000 && bun build/emit.mjs --check && bun run format:check` | pass    | 2027 tests; emit and format clean |
| UX critique          | factory-ux-critic                                                                | not run | no user-facing surface             |

UX critique: skipped

run:run_24733859-559c-469a-9249-64aa85019352

## Post-review

- Reviewer nits (commit 0b460f8b): `rotate_log_file`/`rotate_daemon_logs` now validate `FACTORY_LOG_KEEP` with the same positive-integer guard as `live-stack.sh` (0/non-numeric previously pruned every archive); new test covers `0`, `abc`, `-1`, `3x` rejection and a valid value. The live-writer test spin loop sleeps 10ms per iteration instead of busy-looping.
- Verified: `bun test bin` (167 pass), `bun run format:check`, `bun run check`.

### DIFF
--- a/bin/worktree-daemons.sh
+++ b/bin/worktree-daemons.sh
@@ -5,7 +5,7 @@
 #   bin/worktree-daemons.sh check [WT]        # exit 0 if healthy, non-zero on dead daemon / anomaly
 #   bin/worktree-daemons.sh anomalies [WT]    # print list of anomalies
 #   bin/worktree-daemons.sh supervise [WT]    # watch daemons and restart dead workers with bounded retries
-#   bin/worktree-daemons.sh rotate-logs [WT]  # rotate oversized .factory/run/*.log files
+#   bin/worktree-daemons.sh rotate-logs [WT]  # copy-truncate live daemon logs; rename stopped logs
 #
 set -euo pipefail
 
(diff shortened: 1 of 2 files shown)


hdkiller:
factory-merge-fix round=1 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=768b49076c56dd34fe14f90c5526af221530cb48 new=92d7cf16ccaea553645b0c42b0baff69f25768df

PR #1688, opened 2026-08-30
fix(evals): fail closed on malformed policy subset

### LINKED ISSUE #1108: fix(evals): policy.mjs YAML subset — block scalars absorbed as keys, limits-as-sequence silently defaults
## Problem & Context

Found by cold review of #1094 (implements #1073). The reviewer ran `loadEvalPolicy` over the real `config/policy.yaml`, the real `policy.example.yaml`, the README stanza both appended and inserted mid-file, plus 23 adversarial inputs. It fails closed on almost everything (flow maps, anchors, nested `model:`, quoted `"default"`, missing stanza, bad limits). Three genuine holes remain, all still present on `origin/develop` (verified against `evals/lib/policy.mjs`):

1. **Block-scalar content is absorbed as keys**, contradicting the module docstring, which claims block scalars are "skipped rather than guessed at". In `parseIndentedMap`, a line `note: |` takes the `rest !== ""` branch — it stores the literal string `"|"` and does **not** push a nesting frame, so the following deeper-indented lines are parsed as ordinary keys of the same parent. Reproduced: an `evals:` stanza with `note: |` followed by indented `grader:` / `model: evil-judge` parses to `grader.model = "evil-judge"`. Low real risk (needs an operator to put a block scalar in their own stanza) but the comment asserts a safety property the code lacks. Fix: skip lines indented deeper than a key whose value is `|`/`>` (including the `|-`/`>+` chomping variants), or delete the claim from the docstring.
2. **`limits:` as a YAML sequence silently falls back to defaults.** `parseIndentedMap` drops every `- ...` line via `/^\s*-/`, so `limits:\n  - case_timeout_seconds: 10` yields `{}`; `readLimits` sees a plain object and returns `DEFAULT_LIMITS`, so the runner quietly uses built-in bounds with no error — whereas a flow mapping correctly throws `evals.limits must be a map`. Fix: error when the key is present in the source but yielded no recognized entries.
3. **YAML null accepted as a model name.** `model: null` survives `unquote` as the string `"null"`, which is non-empty and `!== DEFAULT_MODEL`, so it passes `requirePin` and would spawn `claude --model null`. It fails loudly at run time so it degrades honestly, but rejecting `null`/`~` alongside `default` is cheaper.

## Acceptance Criteria

- [ ] Lines indented deeper than a key whose value is a block-scalar indicator (`|`, `>`, and the chomping/indent variants `|-`, `|+`, `>-`, `>+`) are skipped; the reproduction (`note: |` followed by indented `grader:` / `model: evil-judge`) no longer sets `grader.model`, and `loadEvalPolicy` reports the stanza as unpinned. If the implementer instead chooses to drop the safety claim, the docstring in `evals/lib/policy.mjs` must be edited to say block-scalar bodies are read as keys — the code and the comment must agree either way.
- [ ] `limits:` present as a YAML sequence (`limits:\n  - case_timeout_seconds: 10`) throws `EvalConfigError` with message `<file>: evals.limits must be a map` — the same error shape as the existing flow-mapping case — instead of silently defaulting.
- [ ] `evals.grader.model: null` and `evals.grader.model: ~` are rejected as unpinned with the same `problem` shape as `model: default` (a non-null `problem`, `grader === null`, and `requirePin` throwing `EvalConfigError` that includes `REQUIRED_STANZA`).
- [ ] `evals/run.test.mjs` gains one test per hole above; all pre-existing tests in that file (36 on `origin/develop`) still pass.
- [ ] `config/policy.example.yaml` still loads clean through `loadEvalPolicy` (no `problem`) after the parser change — the tightened parser must not reject the shipped example.

## Source File Pointers

- `evals/lib/policy.mjs` — `parseIndentedMap` (block-scalar and sequence handling), `readLimits` (the map check that the sequence case bypasses), `loadEvalPolicy` (grader-model validation, where `null`/`~` must join the `default` rejection), and the module docstring making the "skipped rather than guessed at" claim.
- `evals/run.test.mjs` — existing fail-closed policy tests to extend.
- `config/policy.example.yaml` — the tracked stanza that must keep parsing clean (read-only regression reference; not owned).

## Owned Paths
- evals/lib/policy.mjs
- evals/run.test.mjs

## Verification Command

`bun test --timeout 20000 evals/run.test.mjs`

Confirmed to run clean on `origin/develop` today (36 pass / 0 fail), so a red run after the change is signal, not pre-existing noise. This gate is a unit gate over the parser only: it does not exercise `evals/run.mjs` end to end, so a change that fixes the parser but breaks the runner's consumption of `limits` would not be caught. Keep the `limits` shape returned by `readLimits` unchanged (`caseTimeoutSeconds`, `caseBudgetUsd`, `totalSeconds`, `totalBudgetUsd`) to stay inside what this gate can see.

hdkiller:
Fixes watt-mind/factory#1108

Harden eval policy parsing so malformed limits and grader pins fail closed.

## Validation

| Verification Command | `bun test --timeout 20000 evals/run.test.mjs`    | pass    | 40 tests, 0 failures  |
| Repo verify          | `bun test event-runtime/lib --timeout 20000 && bun build/emit.mjs --check && bun run format:check && bun run lint` | pass    | 0 errors, 615 warnings                  |
| UX critique          | factory-ux-critic                | not run | backend parser and tests only; no user-facing surface |

UX critique: skipped — backend parser and test-only change; no user-facing surface

## Post-review

Reviewer verdict: MERGE with two folds, applied as "honour block-scalar indent indicators" (now 9b0a29aa after the merge-fix loop rebased the branch onto develop):

1. `evals/lib/policy.mjs` — block-scalar header regex now accepts explicit indentation indicators (`|2`, `>2-`, `|-2`) so those bodies are skipped too; test added with `|2` / `>2-`.
2. `evals/lib/policy.mjs` — a `limits:` map with only unrecognized (or no) keys now throws `evals.limits has no recognized keys (expected: …)` instead of `must be a map`; a sequence under `limits:` still reports `must be a map` (the parser now yields a list for it). Test added.

Re-verified: `bun test --timeout 20000 evals/run.test.mjs` (42 pass), `bun run lint` (0 errors), `bun test event-runtime/lib` (0 fail), `bun build/emit.mjs --check`, `bun run format:check`, `bun run check` — all pass.

### DIFF
--- a/evals/lib/policy.mjs
+++ b/evals/lib/policy.mjs
@@ -132,7 +132,20 @@ export function parseIndentedMap(text) {
     ) {
       continue;
     }
-    if (/^\s*-/.test(line)) continue;
+    if (/^\s*-/.test(line)) {
+      // A sequence directly under a map key turns that key into a list, so
+      // callers expecting a map can tell "sequence" apart from "empty map".
+      const top = stack[stack.length - 1];
+      if (
+        top.key !== undefined &&
+        !top.blockScalar &&
+        Object.keys(top.node).length === 0 &&
+        raw.search(/\S|$/) > top.indent
+      ) {
+        top.parent[top.key] = [];
+      }
+      continue;
+    }
     const match = /^(\s*)([A-Za-z0-9_.-]+):\s*(.*)$/.exec(line);
     if (!match) continue;
     const [, pad, key, rest] = match;
@@ -143,10 +156,12 @@ export function parseIndentedMap(text) {
     if (rest.trim() === "") {
       const node = {};
       parent[key] = node;
-      stack.push({ indent, node });
+      stack.push({ indent, node, key, parent });
     } else {
       parent[key] = unquote(rest);
-      if (/^[|>][+-]?$/.test(rest.trim())) {
+      // Block scalar header: `|`/`>` with optional chomping (`+`/`-`) and
+      // explicit indentation (`|2`, `>2-`, `|-2`) indicators, either order.
+      if (/^[|>](?:[+-]?\d?|\d?[+-]?)$/.test(rest.trim())) {
         stack.push({ indent, node: parent, blockScalar: true });
       }
     }
@@ -192,7 +207,9 @@ function readLimits(stanza, file) {
     stanza?.limits !== undefined &&
     !recognizedKeys.some((key) => limits[key] !== undefined)
   ) {
-    throw new EvalConfigError(`${file}: evals.limits must be a map`);
+    throw new EvalConfigError(
+      `${file}: evals.limits has no recognized keys (expected: ${recognizedKeys.join(", ")})`,
+    );
   }
   const pick = (key, fallback) =>
     limits[key] === undefined
--- a/evals/lib/policy.mjs
+++ b/evals/lib/policy.mjs
@@ -126,6 +126,12 @@ export function parseIndentedMap(text) {
   for (const raw of String(text ?? "").split(/\r?\n/)) {
     const line = stripComment(raw);
     if (line.trim() === "") continue;
+    if (
+      stack[stack.length - 1].blockScalar &&
+      raw.search(/\S|$/) > stack[stack.length - 1].indent
+    ) {
+      continue;
+    }
     if (/^\s*-/.test(line)) continue;
     const match = /^(\s*)([A-Za-z0-9_.-]+):\s*(.*)$/.exec(line);
     if (!match) continue;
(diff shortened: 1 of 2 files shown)


hdkiller:
factory-merge-fix round=1 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=18062ce2a895dbb084f3e8eb714c368415f25623 new=9b0a29aa6d96f7fb9d22807b935422c84fbf6f8d

PR #1691, opened 2026-08-30
fix(ask): expose section completeness

### LINKED ISSUE #1110: fix(ask): a partially-failed section reads as authoritative to a JSON consumer
## Problem & Context

Found while reviewing #1069 (PR #1102). Still present on `origin/develop` (verified f6e3cbf4).

`orchestrator/ask.mjs` sets a section's `error` only when **every** repo failed:

```js
error: errors.length === repos.length ? ... : null
```

With two repos configured and one tracker unreachable, `queue.error === null` while `queue.rows` holds only half the answer. The human renderer handles this correctly — `renderRows` appends `partial — <repo>: <err>` for each entry in `section.errors` — but the JSON contract does not: a consumer that reads `error === null` and trusts `rows` gets a silently truncated queue.

This is the same failure mode #1069's AC-4 exists to prevent ("an unreachable tracker must not read as 'no tickets'"), surviving one level down in the machine-readable view — and `ask` was built precisely so that consumers would not each re-derive this.

**Impact.** The three queued consumers (`@factory` Buzz agent, context graph, web Overview) will each read `rows` as complete. With one repo down, the Overview under-reports the queue with no visible signal.

**Chosen fix.** Add `complete: boolean` to every section. The alternative — documenting in the docstring that consumers MUST check `errors[]` as well as `error` — is free but relies on every consumer reading it, and the whole point of the section shape is that a consumer cannot get this wrong by accident.

### Two traps the implementation must avoid

1. **`complete` cannot be a static field set inside `emptySection()`.** Three of the six sections are built empty and mutated afterwards: `doc.recent`, `doc.noop` and `doc.spend` assign `.error` (and `doc.spend` pushes to `.errors`) *after* `emptySection()` returns. A `complete` computed at construction time would be stale and wrong in exactly the failure cases it exists to flag. Derive it once, over every section on `doc`, immediately before `main`/`ask` returns the document (or compute it per section at the point each is finalized) — and add a test that fails if a future section is added without it.
2. **`complete` is not `errors.length === 0`.** The `perRepo` no-repos early return sets `error: "no repositories configured (config/repos.yaml)"` with `errors: []`; so do `doc.recent` and `doc.noop` when there is no runtime database (`error = noDbReason()`, `errors` untouched). Under the literal rule those sections would report `complete: true` next to a set `error`. The rule is `complete === (error === null && errors.length === 0)`.

## Acceptance Criteria

- [ ] Every section object in `ask` JSON output — `queue`, `inflight`, `held`, `recent`, `noop`, `spend` — carries `complete: boolean`, alongside the existing `error` and `errors` fields.
- [ ] `complete` is `true` if and only if the section has `error === null` **and** `errors.length === 0`. Specifically: no repos configured → `complete:false`; no runtime database (`recent`, `noop`, `spend`) → `complete:false`.
- [ ] `complete` reflects the section's final state, not its state at construction: a `spend` section whose `errors` gained a `runtime.db` entry after `emptySection()` reports `complete:false`.
- [ ] The human renderer's output is byte-for-byte unchanged — `unavailable — <err>`, `partial — <repo|source>: <err>`, and the `count()` summary all behave exactly as they do today. `complete` is a JSON-only addition.
- [ ] The module docstring documents `complete` as the field JSON consumers must check, and says why `error` alone is not enough.
- [ ] `orchestrator/ask.test.mjs` covers, over a `perRepo` section: all repos ok → `complete:true`, `error:null`, `errors:[]`; one of two repos failing → `complete:false`, `error:null`, `errors.length===1`, `rows` non-empty; all repos failing → `complete:false` with `error` set.
- [ ] `orchestrator/ask.test.mjs` covers a non-`perRepo` section too: a `--json` document built without a runtime database has `recent.complete === false` and `noop.complete === false`.
- [ ] A test asserts the invariant across the whole document — for every section present in the JSON output, `complete === (error === null && (errors?.length ?? 0) === 0)` — so a section added later cannot silently ship without the field.
- [ ] All pre-existing tests in `orchestrator/ask.test.mjs` (13 on `origin/develop`) still pass.

## Source File Pointers

- `orchestrator/ask.mjs` — `emptySection()` and `perRepo()` (the `errors.length === repos.length` expression and the no-repos early return); the `want.has("queue")` / `"inflight"` / `"held"` blocks that call `perRepo`; the `want.has("recent")` / `"noop"` / `"spend"` blocks that mutate `.error` and `.errors` after construction; `renderRows()` and `count()`, which must keep their current behaviour; and the module docstring where the JSON contract is described.
- `orchestrator/ask.test.mjs` — existing section-shape and error-path tests to extend.
- Context only (not owned): #1069 AC-4, the "unreachable tracker must not read as no tickets" requirement this repeats one level down.

## Owned Paths

- orchestrator/ask.mjs
- orchestrator/ask.test.mjs

## Verification Command

```
bun test --timeout 20000 orchestrator/ask.test.mjs
```

Confirmed to run clean on `origin/develop` today (13 pass / 0 fail), so a red run after the change is real signal. Limit of this gate: it exercises `ask` in-process with stubbed control planes and no real tracker, so it proves the section shape and the `complete` invariant but not the behaviour of the three downstream consumers (Buzz agent, context graph, web Overview). Those are unchanged by this ticket — adding a field is additive — but if any consumer is later taught to *require* `complete`, that is a separate change with its own gate.

hdkiller:
Fixes watt-mind/factory#1110

Consumers can now reject partial ask sections via the final `complete` flag.

## Validation

| Verification Command | `bun test --timeout 20000 orchestrator/ask.test.mjs`    | pass    | 17 tests, 0 failures  |
| Repo verify          | `bun test event-runtime/lib --timeout 20000 && bun build/emit.mjs --check && bun run format:check && bun run lint`  | pass    | clean                  |
| UX critique          | factory-ux-critic                | not run | backend-only JSON contract change; no user-facing surface |

UX critique: skipped — backend-only JSON contract change; no user-facing surface

run:run_37dc785e-4077-403e-bbf9-58d393525a45

### DIFF
--- a/orchestrator/ask.mjs
+++ b/orchestrator/ask.mjs
@@ -28,6 +28,10 @@
  *   never an empty list, while every other section still returns. The command
  *   exits 0 — a partial answer is the point.
  *
+ * JSON consumers must check each section's `complete` field. It is true only
+ * when `error` is null and `errors` is empty; `error === null` alone is not
+ * enough because a section can contain a partial answer and still have rows.
+ *
  * The JSON is the source; the text is rendered from it by `formatAsk`, so the
  * two cannot disagree.
  */
@@ -205,6 +209,17 @@ export function openRuntimeDb(file = dbPath()) {
 }
 
 const emptySection = (extra = {}) => ({ error: null, errors: [], ...extra });
+
+/** Add the final JSON completeness signal after every section has been built. */
+const finalizeSections = (doc) => {
+  for (const name of doc.sections) {
+    const section = doc[name];
+    if (section)
+      section.complete =
+        section.error === null && (section.errors?.length ?? 0) === 0;
+  }
+  return doc;
+};
 
 /**
  * Run `fn` for each repo, collecting rows and per-repo failures.
@@ -589,7 +604,7 @@ export async function gatherAsk({
     if (ownsDb && handle) handle.close();
   }
 
-  return doc;
+  return finalizeSections(doc);
 }
 
 // ---------------------------------------------------------------- render ----
--- a/orchestrator/ask.test.mjs
+++ b/orchestrator/ask.test.mjs
@@ -320,6 +320,11 @@ test("every section is present and typed on a seeded fixture", async () => {
   expect(doc.sections).toEqual([...SECTIONS]);
   for (const name of SECTIONS) expect(doc[name]).toBeDefined();
   expect(doc.repos).toEqual(["factory"]);
+  for (const name of SECTIONS) expect(doc[name].complete).toBe(true);
+  for (const name of SECTIONS)
+    expect(doc[name].complete).toBe(
+      doc[name].error === null && doc[name].errors.length === 0,
+    );
 
   // queue — dispatchable only, as typed rows
   expect(doc.queue.error).toBeNull();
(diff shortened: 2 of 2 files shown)


hdkiller:
factory-merge-fix round=1 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=79381e490972697bb694156142e9cf6c64d1da2e new=33aaca5c81c9ce124b8ab6e7094913aad06372c2

hdkiller:
factory-merge-fix round=2 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=33aaca5c81c9ce124b8ab6e7094913aad06372c2 new=6a6f225c0de134e1896af1f946b9232fc5912b40

Issue #1075, opened 2026-08-29
feat(registry): join usage onto the Agents view — what we dispatch, how often, at what cost, with what outcome


hdkiller:
## Problem & Context

The factory dispatches 53 pinned agent definitions (`event-runtime/agents/*.json`). `event-runtime/web/src/views/Agents.tsx`, backed by `event-runtime/lib/api-registry.mjs`, already lists **what exists** — definitions, event routes, adapters, overrides. What no surface answers is **what we actually dispatch, how often, at what cost, and how well it goes.**

That is the operator question behind Uber's skills marketplace: they run 2,500 skills at 20k executions/day with automated quality checks and continuous feedback from traces. We do not need a marketplace. We need to know which of our 53 agents are load-bearing, which never fire, which fail or refuse disproportionately, and which are quietly expensive — so that skill effort goes where it pays and dead definitions get deleted.

This is joining usage onto the registry that already exists. It is not a new page.

**Known limitation, and the honest way to ship it:** #1056 (*planner RunSpecs do not persist agent defHash pins*) means a run cannot currently be attributed to a *version* of a definition. Attribute by agent **name** now — which is what the interesting questions need — and render the definition-version column as `—` with a one-line note pointing at #1056. Do not block on it, and do not fabricate the attribution.

## Acceptance Criteria

- [ ] `GET /registry/agents` (or a sibling endpoint in `api-registry.mjs`) gains window-scoped usage per agent: `runs`, `failures`, `refusals`, `timeouts`, `p95_execution`, `cost`, `tokens`, `last_run_at`. Reuse the existing aggregation in `event-runtime/lib/metrics.mjs` (`VALID_BREAKDOWN_DIMENSIONS` already includes `agent`) rather than writing a second query layer.
- [ ] The Agents view gains sortable columns for those fields, respecting the existing window selector and `DisplayOptions` machinery — no new page, no new chart library.
- [ ] **Never dispatched** is visible as a first-class state: an agent definition with zero runs in the window renders distinctly (not as `0` lost in a table), because a definition nothing routes to is either dead code or a broken route, and both are findings.
- [ ] Each row's failure/refusal count links to the filtered runs list (`#/runs?agent=…&state=…&window=…`), matching the drill-down pattern #843 establishes for the Metrics page.
- [ ] The definition-version column shows `—` with a tooltip naming #1056 as the reason; it must not display a guessed or run-time-recomputed hash.
[...]

PR #1712, opened 2026-08-30
feat(metrics): add agent refusal and timeout breakdowns

### LINKED ISSUE #1174: feat(metrics): add refusals, timeouts, and last_run_at to the agent breakdown (unblocks #1075)
## Problem & Context
Ticket #1075 requires per-agent registry usage fields (`runs`, failures, refusals, timeouts, p95 execution, cost, tokens, last run time) while requiring reuse of the metrics aggregation. `event-runtime/lib/metrics.mjs` exports agent breakdowns only for runs, failures, cost, tokens, and p95; it does not expose refusals, timeouts, or last-run aggregation. The API response contract types also need to name the added usage shape.

## Evidence
- `metricsBreakdownView()` supports only `runs`, `failures`, `cost`, `tokens`, and `p95_execution`.
- Implementing the remaining required fields outside that module would create the forbidden second aggregation layer.
- #1075 does not own `event-runtime/lib/metrics.mjs` or the web API type contract.

## Acceptance Criteria

- [ ] `VALID_BREAKDOWN_METRICS` in `event-runtime/lib/metrics.mjs` gains `refusals`, `timeouts`, and `last_run_at`. `GET /metrics/breakdown?by=agent&metric=refusals` (likewise `timeouts`, `last_run_at`) returns 200 with rows instead of the current `invalid_metric` error.
- [ ] `refusals` counts runs whose terminal state is `REFUSED` in the window; `timeouts` counts `TIMED_OUT`. Both are window-scoped on `r.updated_at`, the same column the existing `failures` metric uses, so they are strict subsets of it: a fixture with one `REFUSED`, one `TIMED_OUT` and one `FAILED` run for the same agent yields `failures=3`, `refusals=1`, `timeouts=1`.
- [ ] `last_run_at` groups on `r.created_at` and emits, per key, the most recent run start inside the window as `{ key, value, at }` — `value` is epoch milliseconds (so the existing numeric `rows.sort` orders newest first) and `at` is the ISO-8601 string. It is the only metric whose `value` is a timestamp rather than a count/sum, and the JSDoc above `metricsBreakdownView` says so explicitly.
- [ ] The three new metrics are implemented inside `breakdownFacts` and reuse `RUN_FACTS_SELECT` / `RUN_FACTS_JOINS` / `factKeys`, so they compose with every dimension in `VALID_BREAKDOWN_DIMENSIONS` without new joins. `by=modelTier` short-circuits to `modelTierEconomicsView` before `metric` is consumed and stays unchanged.
- [ ] No second aggregation layer is created: `event-runtime/lib/api-registry.mjs` is not modified by this ticket, and the new metrics are reachable only through `metricsBreakdownView`.
- [ ] `event-runtime/web/src/types.ts` names the shape #1075 will consume: `MetricsBreakdownView.rows` gains the optional `at?: string`, and a new exported `RegistryAgentUsage` interface declares `runs`, `failures`, `refusals`, `timeouts`, `p95_execution`, `cost`, `tokens`, `last_run_at` with explicit `| null` where the value can be absent. This is a contract declaration only — no runtime code in this ticket imports it.
- [ ] Tests in `event-runtime/lib/metrics.test.mjs` cover, over a three-agent fixture: refusal and timeout counts distinct from `failures`; `last_run_at` newest-first ordering plus the ISO `at` field; runs outside the window excluded from all three; and `invalid_metric` still thrown for an unknown metric, with the widened list echoed in `validMetrics`.

### Explicit non-goals (do not widen scope)

- **p95 suppression below n=5.** `metricsBreakdownView` currently applies no minimum-sample rule to `p95_execution` (the `samples.length < 5` guard at `metrics.mjs:184` belongs to the percentile-series view). Changing that would alter existing behaviour for `event-runtime/web/src/views/Graph.tsx`, which is not owned here. #1075 already receives `runs` and suppresses on the consumer side.
- **Never-dispatched agents.** A breakdown only returns keys that have facts. Rendering an agent with zero runs is #1075's join against `event-runtime/agents/*.json`.
- **Definition-version attribution.** Blocked on #1056; attribution stays by agent name.

## Source File Pointers

- `event-runtime/lib/metrics.mjs` — `VALID_BREAKDOWN_METRICS` (~line 35), `factKeys` (~412), `breakdownFacts` (~432), `metricsBreakdownView` (~608); the existing `< 5` sample guard at ~184 for reference only.
- `event-runtime/lib/metrics.test.mjs` — existing breakdown coverage (`metric: "p95_execution"` at ~460) to extend.
- `event-runtime/lib/api-metrics.mjs` — `GET /metrics/breakdown` passes `metric` straight through; validation lives in `metrics.mjs`, so no change is expected here.
- `event-runtime/web/src/types.ts` — `MetricsBreakdownView` (~277).
- `event-runtime/lib/db.mjs` — `runs.state`, `runs.created_at`, `runs.updated_at`, `run_usage`.
- `event-runtime/lib/api-registry.mjs` — the consumer in #1075. Read for context; do not edit.

## Owned Paths

- event-runtime/lib/metrics.mjs
- event-runtime/lib/metrics.test.mjs
- event-runtime/web/src/types.ts

## Verification Command

```
bun test --timeout 20000 event-runtime/lib/metrics.test.mjs event-runtime/lib/api-metrics.test.mjs && bun run lint
```

**Gate limitation:** the root gate has no `tsc --noEmit` step (`bun run lint` is eslint only; typechecking lives in `event-runtime/web`'s own `build` script). The `types.ts` addition is therefore lint-checked but not structurally typechecked by the command above. If the web deps are installed, additionally run `cd event-runtime/web && bunx tsc --noEmit` before opening the PR.

## Relations

- Unblocks #1075 (`ai:blocked`). #1075 is blocked *on this ticket*, not the reverse — nothing here waits on a human decision or on #1075's output.
- Related: #1056 (planner RunSpecs do not persist agent defHash pins), which is why attribution is by agent name.

hdkiller:
Fixes watt-mind/factory#1174

Expose refusal, timeout, and newest-run agent metrics for #1075.

## Validation

| Check                | Command                                                                                          | Result  | Notes                              |
| -------------------- | ------------------------------------------------------------------------------------------------ | ------- | ---------------------------------- |
| Verification Command | `bun test --timeout 20000 event-runtime/lib/metrics.test.mjs event-runtime/lib/api-metrics.test.mjs && bun run lint` | pass    | exit 0; tests and lint completed  |
| Repo verify          | `bun test event-runtime/lib --timeout 20000 && bun build/emit.mjs --check && bun run format:check && bun run lint` | pass    | exit 0; clean                     |
| Web TypeScript       | `cd event-runtime/web && bun x tsc --noEmit`                                                    | pass    | no diagnostics                     |
| UX critique          | factory-ux-critic                                                                                | not run | skipped; no user-facing surface   |

UX critique: skipped — shared metrics and type-contract changes only

run:run_e9c47bc4-0bc6-4049-a9b2-4a15dffe2ffc

## Post-review

- Reviewer fold (a1915ac8): `metricsBreakdownView` `last_run_at` reduction now skips `created_at` values where `!Number.isFinite(Date.parse(...))`, so an unparsable timestamp seen first for a key can no longer pin that key to `NaN`. Covered by a `delta@1` fixture in `metrics.test.mjs` (fails without the guard, passes with it).
- Verified: ticket Verification Command, `bun run lint`, `bun test event-runtime/lib` (2143 pass), `bun build/emit.mjs --check`, `bun run format:check`, `bun run check`.

### DIFF
--- a/event-runtime/lib/metrics.mjs
+++ b/event-runtime/lib/metrics.mjs
@@ -35,9 +35,12 @@ export const VALID_BREAKDOWN_DIMENSIONS = Object.freeze([
 export const VALID_BREAKDOWN_METRICS = Object.freeze([
   "runs",
   "failures",
+  "refusals",
+  "timeouts",
   "cost",
   "tokens",
   "p95_execution",
+  "last_run_at",
 ]);
 
 export const MAX_METRIC_BUCKETS = 500;
@@ -431,17 +434,33 @@ function factKeys(row, dimension) {
 }
 
 function breakdownFacts(db, metric, range) {
-  if (metric === "runs" || metric === "failures") {
-    const failures =
-      metric === "failures"
-        ? "AND r.state IN ('FAILED', 'REFUSED', 'TIMED_OUT', 'CANCELLED')"
-        : "";
-    const time = metric === "failures" ? "r.updated_at" : "r.created_at";
+  if (
+    metric === "runs" ||
+    metric === "failures" ||
+    metric === "refusals" ||
+    metric === "timeouts"
+  ) {
+    const states = {
+      failures: "'FAILED', 'REFUSED', 'TIMED_OUT', 'CANCELLED'",
+      refusals: "'REFUSED'",
+      timeouts: "'TIMED_OUT'",
+    };
+    const state = states[metric];
+    const time = state ? "r.updated_at" : "r.created_at";
     return db
       .query(
         `${RUN_FACTS_SELECT}, 1 AS value
        FROM runs r ${RUN_FACTS_JOINS}
-       WHERE ${time} >= ? AND ${time} < ? ${failures}`,
+       WHERE ${time} >= ? AND ${time} < ? ${state ? `AND r.state IN (${state})` : ""}`,
+      )
+      .all(range.startIso, range.endIso);
+  }
+  if (metric === "last_run_at") {
+    return db
+      .query(
+        `${RUN_FACTS_SELECT}, r.created_at AS last_run_at
+       FROM runs r ${RUN_FACTS_JOINS}
+       WHERE r.created_at >= ? AND r.created_at < ?`,
       )
       .all(range.startIso, range.endIso);
   }
@@ -604,7 +623,11 @@ export function modelTierEconomicsView(
     .sort((a, b) => a.key.localeCompare(b.key));
 }
 
-/** Top-N operational dimensions used by GET /metrics/breakdown. */
+/**
+ * Top-N operational dimensions used by GET /metrics/breakdown. `last_run_at`
+ * is the only timestamp metric: its row value is epoch milliseconds, with the
+ * corresponding ISO-8601 timestamp exposed as `at`.
+ */
 export function metricsBreakdownView(
   db,
   { now, window = "24h", by, metric, limit } = {},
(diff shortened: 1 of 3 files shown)


hdkiller:
factory-merge-fix round=1 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=a931bd2859fc8bd2d09ff5e3b056ead8b5ce6ab7 new=a8c780a827f8c89f8a51e4e638de65a61a5e2ea6

PR #1751, opened 2026-08-30
fix(planner): deduplicate unchanged human-needed proposals

### LINKED ISSUE #1730: fix(planner): do not re-propose owned_paths_not_closed human_needed for an unchanged ticket body
## Context
work-scan proposes tickets whose `## Owned Paths` are not closed on every scan (~10 min), producing a fresh `human_needed` proposal each time — e.g. #1219 produced three `owned_paths_not_closed` proposals in 15 minutes, so the inbox fills with identical rows for a ticket nothing has changed on.

`event-runtime/lib/planner.mjs:1543-1553` returns `refusal("owned_paths_not_closed", evidence, "human_needed")`; the dispatch-gate branch at planner.mjs:2424-2426 calls `humanNeeded()` (planner.mjs:2070-2081), which unconditionally `insertProposal`s a new open proposal. There is no lookup of an already-open `human_needed` proposal for the same repo/ticket/reason, and `factory.dispatch.requested`'s idempotency key (event-types.json:170-175, `["inputHash"]`) is only consulted at run insertion, which this refusal never reaches (and the key is suffixed with the chain correlation at planner.mjs:366-373, so each scan differs anyway). The evidence already carries `ticket.descriptionHash` (planner.mjs:1515-1523), so "body unchanged" is cheaply detectable.

## Acceptance criteria
- Before `humanNeeded()` inserts, the planner looks for an open proposal with `decision = 'human_needed'`, the same reason prefix, the same repo+ticket, and the same ticket body hash (store the hash in the proposal evidence/reason detail if it is not already queryable). When one exists, the new event resolves `noop` with reason `human_needed_already_open:<proposalId>` and no new proposal row.
- A ticket whose body hash changed (new ready-pin) produces a new `human_needed` proposal and the older one is superseded (reuse the `superseded` status from planner.mjs:2754).
- Approving/dismissing/expiring the open proposal re-enables a fresh proposal on the next scan.
- Tests in planner.test.mjs: (a) two consecutive dispatch events for the same unclosed ticket → one open proposal, second event `noop` with the typed reason; (b) body change → second proposal, first superseded; (c) the guard applies to `owned_paths_not_closed` and the other `human_needed` ticket refusals routed through the dispatch gate, not to `noop` refusals.

## Owned Paths
- event-runtime/lib/planner.mjs
- event-runtime/lib/planner.test.mjs

## Verification Command
```
bun test event-runtime/lib/planner.test.mjs --timeout 60000
```

hdkiller:
Fixes watt-mind/factory#1730

Suppress duplicate unchanged-ticket human-needed proposals while retaining fresh prompts for changed bodies.

## Validation

| Verification Command | `bun test event-runtime/lib/planner.test.mjs --timeout 60000` | pass | 104 tests, 0 failures |
| Repo verify          | `bun test event-runtime/lib --timeout 20000 && bun build/emit.mjs --check && bun run format:check && bun run lint` | pass | 2178 pass, 10 skipped |
| UX critique          | factory-ux-critic                | not run | no user-facing surface |

UX critique: skipped

run:run_a0dfcaca-7a0d-412b-aff9-3be9ff173e00

## Post-review

- The `humanNeeded` dedupe guard now treats an open proposal as current only while it is within its TTL (`created_at + ttl_seconds`, computed at plan time `at`), matching how `proposals.mjs`, `janitor.mjs` and `metrics.mjs` derive expiry. Production never writes `status='expired'`, so previously an unchanged ticket was never re-proposed after its question aged out (AC3). Aged-out rows are superseded like stale ones.
- Supersede reason is `superseded_by_ticket_body_change` only when the prior row carried a body-hash marker that differs from the current hash; otherwise `superseded_by_replan`.
- Test no longer fakes expiry via `UPDATE proposals SET status='expired'`; it plans at `NOW + ttl_seconds` (still noop) and `NOW + ttl_seconds + 1s` (fresh proposal, old row `superseded_by_replan`), and asserts the body-change supersede reason.
- The `[dispatch_ticket_body_hash:…]` marker stays in `proposals.reason`: `spec_json`/`spec_hash` are not a cheap alternative because a non-null `spec_json` on a proposal routes it through the approve/run-spec paths in `proposals.mjs` (`recordedSpec` / `storedSpec`) and auto-approval hash checks. Human-facing views (inbox/web) are outside this PR's owned paths, so the marker is not stripped there; a follow-up can strip it in the renderers if it proves noisy.

### DIFF
--- a/event-runtime/lib/planner.mjs
+++ b/event-runtime/lib/planner.mjs
@@ -2106,7 +2106,73 @@ function coalesceWebhookMergeRequest(
   };
 }
 
-function humanNeeded(db, event, reason, at, ttlSeconds) {
+function humanNeededReasonPrefix(reason) {
+  return String(reason).split(":", 1)[0];
+}
+
+function ticketHumanNeededContext(payload, evidence) {
+  const repo = payload?.repo;
+  const ticket = payload?.ticket;
+  const descriptionHash = evidence?.ticket?.descriptionHash;
+  // A missing ticket has no body to compare. Its synthetic empty-description
+  // hash must not make unrelated failed lookups suppress each other.
+  if (
+    evidence?.checks?.ticket_found !== true ||
+    typeof repo !== "string" ||
+    typeof ticket !== "string" ||
+    typeof descriptionHash !== "string"
+  ) {
+    return null;
+  }
+  return { repo, ticket, descriptionHash };
+}
+
+function humanNeeded(db, event, reason, at, ttlSeconds, ticketContext = null) {
+  const outcomeReason = reason;
+  if (ticketContext) {
+    const reasonPrefix = humanNeededReasonPrefix(reason);
+    const matching = db
+      .query(
+        `SELECT p.*
+         FROM proposals p
+         JOIN events e
+           ON e.source = p.event_source AND e.event_id = p.event_id
+         WHERE p.decision = 'human_needed'
+           AND p.status = 'open'
+           AND json_extract(e.envelope_json, '$.payload.repo') = ?
+           AND json_extract(e.envelope_json, '$.payload.ticket') = ?
+           AND substr(p.reason, 1, ?) = ?
+         ORDER BY p.created_at, p.rowid`,
+      )
+      .all(
+        ticketContext.repo,
+        ticketContext.ticket,
+        reasonPrefix.length,
+        reasonPrefix,
+      );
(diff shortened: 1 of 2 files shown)


hdkiller:
factory-merge-fix round=1 finding=6e27a6319dc7dcc61478b9b5214e7390843aa7210328f7299f0768394922ae62 old=ecdcf713d60efa855599dfe4c02c39c8afb69182 new=d89255d0969942fb1af5ee9c669204c28d9af024

PR #2044, opened 2026-08-31
fix(ci): harden sandbox smoke gates (#2042)

### LINKED ISSUE #2042: fix(ci): sandboxed-verify smoke gates merges on the quarantined timing group and mishandles sandbox_unavailable
## Summary
The new merge-gating `Sandboxed verify smoke` lane (PR #2037) runs `bun test event-runtime/lib --timeout 20000` WITHOUT the WM-918 timing-group exclusion that `Fast unit tests` (~ci.yml:268) and `Full verification` (~:603) both apply — the dedicated `Timing-bound tests` lane runs that group non-required with `--retry 2 --max-concurrency=2` precisely because it flakes under shared-host contention. The smoke therefore gates every future PR on a group the repo already declared too flaky to gate merges, at default concurrency, inside a 1 GiB network-less namespace. It passed its first run (92 s); this is a standing risk, not an observed failure. Also from the same review: `SandboxUnavailable` is uncaught in the inline runner script (a host without unprivileged userns dies with a bare stack trace; the cloud lane — fork PRs, dependabot, nightly schedule — has never executed the job), and `handoffSandboxFacts()` hardcodes outer-mode facts that are wrong for the nested branch while the new test pins the hardcoding.

## Acceptance Criteria
- [ ] The smoke applies the same WM-918 timing exclusion as Fast unit tests / Full verification, OR retries the timing names, OR the accepted flake policy is measured and documented in the workflow comment — pick one and say why.
- [ ] The inline runner wraps runHandoffVerifySmoke: on reasonCode `sandbox_unavailable` it emits `::error::sandbox_unavailable on <lane> runner` (self-hosted) or a documented explicit outcome on the cloud lane, never a bare stack trace; a test covers the wrapper mapping.
- [ ] `handoffSandboxFacts()` derives HOME/workspace facts from the observation's confinement string (correct in nested mode); test updated to stop pinning the hardcoded string.
- [ ] Optional (say if skipped): gate the full-suite output dump on failure only, keeping the facts line.

## Owned Paths
- .github/workflows/ci.yml
- event-runtime/lib/verify.mjs
- event-runtime/lib/verify.test.mjs

## Verification Command
`FACTORY_EVENT_HOME=$(mktemp -d) FACTORY_LINEAR_OFFLINE=1 bun test event-runtime/lib/verify.test.mjs && bun run lint`

hdkiller:
Fixes watt-mind/factory#2042

Keeps the required sandbox smoke deterministic and makes unavailable isolation explicit.

## Validation

| Check                | Command                                                                                                  | Result  | Notes                         |
| -------------------- | -------------------------------------------------------------------------------------------------------- | ------- | ----------------------------- |
| Verification Command | `FACTORY_EVENT_HOME=$(mktemp -d) FACTORY_LINEAR_OFFLINE=1 bun test event-runtime/lib/verify.test.mjs && bun run lint` | pass | tests and lint completed |
| Repo verify          | `bun test event-runtime/lib --timeout 20000 && bun build/emit.mjs --check && bun run format:check && bun run lint` | pass | all stages completed |
| UX critique          | `factory-ux-critic`                                                                                      | not run | no user-facing surface        |

UX critique: skipped

run:run_d4c96527-937e-4fe8-a11e-c3e36f50da7c

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -466,12 +466,13 @@ jobs:
         # WM-918: required lanes exclude the timing-bound group because its
         # wall-clock ceilings are not reliable under shared-host contention.
         env:
-          FACTORY_HANDOFF_SMOKE_COMMAND: >-
-            bun test event-runtime/lib --timeout 20000 --test-name-pattern "$(bun event-runtime/lib/test-helpers-timing.mjs exclude-pattern)" && bun build/emit.mjs --check && bun run format:check && bun run lint
           FACTORY_HANDOFF_SMOKE_LANE: ${{ runner.environment }}
         shell: bash
         run: |
           set -euo pipefail
+          pat="$(bun event-runtime/lib/test-helpers-timing.mjs exclude-pattern)"
+          [ -n "$pat" ] || exit 1
+          export FACTORY_HANDOFF_SMOKE_COMMAND="bun test event-runtime/lib --timeout 20000 --test-name-pattern \"$pat\" && bun build/emit.mjs --check && bun run format:check && bun run lint"
           bun -e '
             import { mkdtempSync } from "node:fs";
             import { tmpdir } from "node:os";
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -463,8 +463,12 @@ jobs:
         # config/repos.yaml. `runHandoffVerifySmoke` delegates to
         # `runHandoffCommand`, so unshare, env scrubbing, the worktree-only
         # mount, chroot, and HOME=/tmp/home all stay owned by production code.
+        # WM-918: required lanes exclude the timing-bound group because its
+        # wall-clock ceilings are not reliable under shared-host contention.
         env:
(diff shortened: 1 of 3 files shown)


hdkiller:
## Post-review fixes

- `.github/workflows/ci.yml`: `FACTORY_HANDOFF_SMOKE_COMMAND` embedded `$(bun event-runtime/lib/test-helpers-timing.mjs exclude-pattern)` for evaluation by the **sandbox** bash. If the helper fails inside the sandbox, `$()` substitutes empty and `--test-name-pattern ""` matches everything, silently re-admitting the timing group into the required gate. Fixed by computing `pat="$(bun event-runtime/lib/test-helpers-timing.mjs exclude-pattern)"` in the runner's own `run:` block under `set -euo pipefail` (with `[ -n "$pat" ] || exit 1`), then interpolating the literal pattern into the exported command string — matching the pattern already used in the other lanes (ci.yml:223/:517).

Verified: `bun run lint` clean (0 errors), `actionlint .github/workflows/ci.yml` clean. Pushed as 96b1eb1e.

PR #2337, opened 2026-09-15
feat(ship): pre-flight and chain for develop→master releases (WM-1104)

hdkiller:
Fixes WM-1104

Shipping legalease on 2026-09-15 took ~2.5 h of an agent re-deriving, from `ci.yml`, `verify_case_agent_release.py` and memory, a sequence that is entirely mechanical. This turns it into two commands.

## `factory ship-preflight --repo <name> [--json] [--no-fetch]`

Read-only. One PASS/FAIL/SKIP line per check, the exact next command on every FAIL, exit 0 only when nothing failed.

- **base-green** — every job of the newest CI run *for the exact tip SHA*. Newest by run number then attempt, so a superseded (cancelled) run is not mistaken for red.
- **runtime-pin:&lt;role&gt;** — the pinned source commit is an ancestor of the tip **and** nothing under that publisher's own `on.push.paths` moved since. FAIL prints `gh workflow run <publisher> --ref <base> -f commit_sha=<tip>`.
- **publishers-idle**, **pin-pr**, **escalated-pr**, **ssh-probe** (promotion journal must print exactly `clean`).

An unreachable forge, a failed `git`, or a probe that errors is a FAIL — it fails closed.

## `factory ship-chain --repo <name> --until <preflight|pin|release> [--apply]`

**Dry run unless `--apply`**, and a dry run is a plan, not a rehearsal: it makes no mutating call and does no waiting. Per round it merges an already-open pin PR first (that merge moves the tip), otherwise dispatches stale roles' publishers — one at a time under `serial_publishers`, because they share one promotion lock — waits for the publisher, waits for the reconcile PR, merges it, and re-runs pre-flight. `--until release` then opens `base → deploy_branch`, merges with `--merge`, and runs `post_release_checks`.

**Merges gate on the check-run summary, never on a watched run's exit status** — `gh run watch` can exit 0 for one workflow while another check run on the same head is red.

## Three judgement calls worth reviewing

1. **Skipped is not failed.** A literal "every non-advisory job must be `success`" is unpassable: `deploy-prod` is always skipped on develop. A job skipped because its *dependency* failed is still caught, through the dependency.
2. **The forge contract grew three verbs** (`workflowDispatch`, `prCreate`, `prMerge`) instead of `ship.mjs` spawning `gh`, per `lib/forge/types.mjs`' rule that nothing outside `lib/forge/` speaks `gh`. No `runWatch` verb: waiting is bounded polling of existing read verbs at ≥60 s, which the floor's rate-limit rule already requires. Both implementations and the shared contract suite cover them.
3. **`config/repos.yaml` is gitignored**, and the tracked example holds no client repos by policy (its own header). So the schema is documented on the example's commented entry and in `docs/ship.md`; legalease's real `runtime_roles` / `ssh_probe` / `post_release_checks` live in the operator-local file, which is what the live run below exercises. Nothing machine- or client-specific is committed.

A role built in another repository (legalease pins lawz's research runner) reports **SKIP** with the reason rather than answering an ancestry question about a history this repo does not contain.

## Verification

`bun test orchestrator/ship.test.mjs lib/forge/contract.test.mjs` — 82 pass, 0 fail (33 ship + 49 forge contract). Fixtures cover pin stale, pin fresh, pin off-branch, foreign-source skip, publisher running, bot pin PR open, escalated PR, ssh dirty/unconfigured, base red, base superseded, advisory red ignored, skipped-not-failed, unreachable forge, dry run makes no mutating call, and a red check run refusing the merge.

`bun run lint` — 0 errors, 616 warnings (the baseline; this branch adds none). `bun run check` — exit 0 (emit drift, web typecheck, prettier).

Live read-only run from the worktree against the real repos:

```
$ factory ship-preflight --repo legalease

ship pre-flight: legalease (develop @ 4f236ccc)

  PASS  develop tip fully green            run 34951809069: 10 success, 4 skipped, 0 advisory
  PASS  runtime pin case_agent fresh       pinned 7529fd33, no publisher path changed since
  SKIP  runtime pin research_runner fresh  built from watt-mind/lawz; its pinned commit is not in this history, so freshness is that repo's publisher to prove
  PASS  no publisher run in progress       2 publisher workflow(s) idle
  PASS  no open bot pin PR                 none by app/watt-mind-factory
  PASS  no escalated PR targeting develop  none
  PASS  promotion journal clean            clean

PASS — 7 checks, nothing blocking. The release PR may be opened.
EXIT=0
```

`factory ship-chain --repo legalease --until release` (dry run) plans exactly `open-release-pr`, `merge-release-pr`, and the three `post-release-check` commands, and calls nothing.

## Deliberately not in this PR

- **`--apply` has not been exercised against a live repository.** It is fixture-tested end to end, but the first real use should be an operator watching it.
- **No `runWatch` forge verb** and no reuse of `gh run watch`; waiting is polling.
- **`research_runner` freshness is not computed** — it is SKIP by design until lawz's side can be asked.
- **No scheduler wiring.** `ship-preflight` is not added to `tick`/`watchdog`; it is a command `/factory-ship` calls.


hdkiller:
## Fix round 1 — all 11 ESCALATE findings closed

Head: `e20d0b0b`. Every finding below has a regression test that fails on `e1b4523b`.

`bun test orchestrator/ship.test.mjs lib/forge/contract.test.mjs` — **114 pass, 0 fail** (62 ship, up from 33; 52 forge contract, up from 49). `bun run lint` 0 errors, `bun run check` exit 0, prettier clean.

### 1. TOCTOU on the master merge

Everything above the release merge was decided against a pre-flight that could be many minutes old, and the merge itself named no commit.

- `preflight()` is **re-run immediately before** the release merge; any FAIL throws `pre-flight went red between the plan and the release merge: <check>: <detail>`.
- `mergeWhenGreen` gained `requireHead`: the release PR's head must be the pre-flighted tip, both before the wait and after it.
- After the check-run wait, the PR is **re-read** and a moved head is refused — an in-process check that closes most of the window.
- New Forge option `prMerge(..., { matchHeadCommit })` emits `gh pr merge --match-head-commit <sha>` (github), refuses a mismatched head (memory), and is used for **both** the pin-PR merge and the release merge. That is what closes the rest of the window: only GitHub can enforce it at the instant of the merge.

Tests: `prMerge --match-head-commit refuses a head that moved, and merges the one it named` (contract suite, both implementations), `prMerge emits --match-head-commit only when the caller names a SHA` (arg emission), `TOCTOU: a tip that moved between the plan and the merge is refused`, `TOCTOU: a head that moves while the check runs are waited on is refused`, `a base that goes red between the plan and the merge stops the chain`, plus the merge-op assertions in the happy-path and pin-PR tests.

### 2. `checkBaseGreen` ignored `run.conclusion` and treated zero jobs as green

`base-green` now reads the run's own conclusion **before** its jobs (`!== "success"` → FAIL), and additionally FAILs when no non-advisory job succeeded. A `startup_failure` is a *completed* run with an empty job list, and an empty job list satisfies every per-job test — the org's documented zero-job red run.

Tests: `base red by conclusion: a failed run is a FAIL even when its jobs read green`, `base zero-job: a startup_failure run with no jobs is a FAIL, not a green tip` (covers both the `startup_failure` conclusion and a nominally-`success` run with zero jobs), `a zero-job base run fails pre-flight and never reaches the release PR`.

### 3. `checkRunsGreen([])` returned ok

`[].every(green)` is `true`, so a commit whose checks did not exist yet satisfied the merge gate instantly. It now requires at least one **completed, non-skipped, successful** check run (`skipped` alone is not evidence anything ran) and returns a `missing` reason.

On "the configured CI workflow's check is among them" — I implemented the requirement but not by name-matching, and want this called out: GitHub names a check run after the **job**, not the workflow, so on this repo `cfg.ciWorkflow` is `CI` while the check-run names are `Verify`, `Fast unit tests`, … A literal name match would fail on real data. Instead `mergeWhenGreen` proves it from the Actions runs API for the same head SHA — `selectRun(runsForSha(sha), { workflow: cfg.ciWorkflow })` must exist, be `completed`, and have concluded `success`. Same guarantee, real evidence. Happy to switch to name-matching if you meant something narrower.

Tests: `checkRunsGreen refuses an empty list, and a list that is only skips`, `empty check runs never satisfy the merge gate` (through `--until release --apply`; the wait now times out instead of merging), `green check runs on a head the CI workflow never ran for are refused`.

### 4. Escalated / draft release PR

`isEscalatedPr` only inspects PRs targeting `base`, so the release PR — which targets `deployBranch` — was structurally invisible to the gate. New pure `releasePrRefusal(cfg, pr)` checks the `escalated` label, `isDraft`, that the head really is `cfg.base` and the target really is `cfg.deployBranch` (that last one also catches reusing some other branch's PR onto master). `shipChain` refuses before `mergeWhenGreen`. `isDraft` added to `PR_FIELDS`; `hasEscalatedLabel` factored out.

Tests: `releasePrRefusal names why a release PR must not be merged` (all five refusal shapes), `an escalated release PR is refused, not merged`, `a draft release PR is refused, not merged`.

### 5. Role with missing/misspelled `paths` yielded SKIP

A check with no definition of stale can never fail — and `path:` instead of `paths:` turned it into a quiet yellow line on every release. `loadShipConfig` now refuses a non-foreign role without `paths`, naming the role index and its publisher workflow and pointing at `source_repo` for the legitimate case. `checkRuntimeRole` FAILs rather than SKIPs if a hand-built config reaches it anyway.

Tests: `a locally-built role with no paths is refused at load, by name` (uses the `path:` typo), `a foreign role needs no paths, and carries its optional foreign_repo`, `a locally-built role with no paths is a FAIL, never a SKIP`.

### 6. `activeRuns` missed `waiting` / `requested`

Filtering the repo-wide run list by `status=in_progress` + `status=queued` misses `waiting` (an environment approval), `requested` and `pending`. Replaced with `workflowRuns(repo, workflow)` — the workflow's *own* runs, newest first — and `isActiveRun = status !== "completed"`. No status vocabulary to keep in sync with GitHub.

Tests: `test.each(["waiting", "requested", "pending"])` — `a publisher run in %s status is not idle`, plus `a completed publisher run leaves the publishers idle` and the existing in-progress case.

### 7. Publisher wait fired before the dispatched run existed, and never read its conclusion

`workflow_dispatch` returns no run id and the run does not exist when the call returns, so "no active run of this workflow" was satisfied *before the publisher started* — the chain then waited 45 min for a pin PR the publisher had not begun to earn. And a publisher that failed produced no pin, which the chain also never noticed.

Now: snapshot the workflow's run ids → dispatch → poll (bounded at 10 min, `DISPATCH_APPEAR_TIMEOUT_MS`) for a run that is **new** and not older than the dispatch by more than a 2-min clock skew → poll until it is `completed` → assert `conclusion === "success"`, else `ShipChainError: publisher <wf> run <id> concluded <x> — the pin it would have produced does not exist` with `gh run view --log-failed` as `next`. A `publisher-succeeded` step is recorded.

Tests: `a publisher whose run never appears times out on the dispatch, not on the pin PR`, `a failed publisher run stops the chain by name instead of waiting for a pin PR`, `a publisher still in progress is waited on before the pin PR is looked for` (the fixture only goes green on the fourth poll, so a finish-wait iteration genuinely observes `in_progress`), and the happy path asserts `publisher-succeeded` in the plan.

### 8. `research_runner` SKIP printed nothing

A SKIP with no detail is indistinguishable from a check that never ran. The manifest is now read **first**, so the line always prints `pinned <sha8> (source watt-mind/lawz)`; a manifest with no commit for the role is a FAIL as it is for local roles. New optional per-role `foreign_repo` (deliberately separate from `source_repo`, since it costs a read against a repo the operator may not have access to) makes it also read that repo's newest successful publisher run — `repos/<repo>/actions/workflows/<wf>/runs?status=success&per_page=1`, the API form of `gh run list -R … --status success --limit 1` — and downgrade the SKIP to **WARN** when the head differs, with that `gh run list` as `next`. WARN never blocks: `report.ok` is still "no check failed".

Tests: `a foreign-source role is skipped, but names the pinned commit and its source`, `a foreign role with foreign_repo WARNs when the pin is behind that repo's publisher` (also asserts `report.ok === true` and that `formatPreflight` renders `WARN`), `a foreign role whose pin matches that repo's newest publisher run stays a SKIP`. `config/repos.yaml` (operator-local, gitignored) has `foreign_repo: watt-mind/lawz` wired for legalease; the shape is documented on `config/repos.example.yaml` and in `docs/ship.md`.

### 9. Docs step 6 could not work

New `--until post-release`: runs **only** `post_release_checks` against the live deploy — no pre-flight, no PR, no merge. It runs with or without `--apply` because these are the repo's own read-only verification commands and there is nothing to dry-run; `--apply` adds the notification. A repo with no `post_release_checks` is a `ShipConfigError` rather than a silent success. `shared/commands/factory-ship.md` step 6 now points at it (pins re-run, `bun build/emit.mjs` regenerated `plugins/core` + `dist`), as do `docs/ship.md` and `bin/factory`'s help.

Because `report` is `null` in this mode, `shipChain` now returns `repo: cfg.name` on every path, `formatChain` reads that instead of `result.report.repo`, and the CLI skips `formatPreflight` when there is no report.

Tests: `runs only the post-release checks — no pre-flight, no PR, no merge` (asserts no `prCreate`/`prMerge`/`workflowDispatch` at all), `a red check exits non-zero, and notifies only under --apply`, `a repo with no post_release_checks says so instead of reporting success`.

### 10. Red post-release check wording + notify

A red post-release check is not a plan that stopped — the release is already on the deploy branch. `formatChain` now prints `POST-RELEASE CHECK FAILED: <command>`, the line `the release is merged — this is a live deploy problem, not a plan that stopped`, the last line of the command's output, and a footer `RELEASE MERGED, N POST-RELEASE CHECK(S) RED — treat as an outage`. Under `--apply` an injectable `notify` dep is called with `SMOKE RED <repo>: <command> failed after release <sha>`; `defaultDeps` wires it to `bin/factory notify` with the message as a single argument (never a flag).

Tests: `a red post-release check exits non-zero, says the release is merged, and notifies` asserts the exact notify payload, that the merge did happen first, and all three output strings.

### 11. Tests for the apply path

New `describe("shipChain --until release --apply")` with eight fixture-driven cases: happy path (`prCreate` → re-run pre-flight → `mergeWhenGreen` with `matchHeadCommit` → post-release checks, asserting the plan and that nothing was notified), tip-moved TOCTOU, head-moved-during-wait TOCTOU, base-went-red TOCTOU, escalated release PR, draft release PR, empty check runs, missing CI run for the head, zero-job base run, and the red post-release check. Plus a `--until post-release` block.

The fixture forge grew `/actions/workflows/<wf>/runs` (with `status=success` and `per_page`), a per-head-SHA `runsBySha` override, and `isDraft` on PRs; `fakeGit` accepts a function for `tip` so a moving base can be simulated.

---

### Also filed (review polish items 12-14, not in this PR)

- #2340 — no pagination on the four `per_page=100` reads; every truncation biases toward green.
- #2341 — `runtime_roles.paths` duplicates the publisher workflow's `on.push.paths`; read the workflow or drift-check it.
- #2342 — GitHub negated globs (`!path`) passed to git as literal pathspecs.

### CI

`CI` run 34957749666 green. Check runs on `e20d0b0b`: 14 success, 2 skipped (`Playwright smoke journeys`, `Pinned tools forced-install smoke`), 0 failure — including both required checks (`Verify`, `Shadow runner fleet available`). `CodeQL` went red on the first attempt during its SARIF **upload** step (develop's tip shows the same failure); re-run once and green.

### One local note

`orchestrator/repos-config.test.mjs` fails on this machine against the operator-local `config/repos.yaml`, whose `factory.verify` lacks the ` && bun run format:check` the test asserts. Pre-existing and unrelated to this branch — CI reads `config/repos.example.yaml` and is green.

PR #2339, opened 2026-09-15
feat(infra): recovery verbs for digest rotation, stranded transactions and evicted pods (WM-1105)

hdkiller:
Fixes WM-1105

Three documented legal-dev recoveries, each of which was blocked by the coding agent's permission classifier during the 2026-09-15 release, wrapped as `factory infra <verb>` so the verbs carry their own preconditions and audit log; a blanket `Bash(factory infra *)` allow-rule is deliberately NOT recommended (prefix rules cannot see `--yes`) instead of `ssh … sudo rm -rf …`.

## What landed

- `orchestrator/infra.mjs` — `rotate-driver-digest` (OPS-694), `discard-unprepared-transaction` (CLNT-3170), `sweep-evicted-pods` (CLNT-3167). Pure `evaluate*` / `select*` decision functions with `ssh`, `gh`, `kubectl`, deployment-resourceVersion, clock, log sink and ticket-comment effects injected through a trailing options object, per `docs/event-runtime-conventions.md`. Exit codes follow `branch-guard.mjs`: 0 clean / 2 REFUSED / 3 CANNOT EVALUATE.
- `orchestrator/infra.test.mjs` — 56 tests, entirely on fake adapter output.
- `bin/factory` — `infra)` dispatch plus the help-list entry.
- `docs/infra-recovery.md` — the three verbs, the incident behind each, and the `Bash(factory infra *)` allow-rule to paste into `~/.claude/settings.json`.

## Safety model

Dry run is the default for all three; `--yes` is required to act and `--dry-run` is an explicit spelling of the default. Every invocation tees to `~/.factory/logs/infra/<verb>-<stamp>.log`; `--ticket <ID>` posts one line.

The refusals are the point:

- **digest rotation** refuses unless `--digest` equals *both* `sha256sum` of the installed driver *and* `driver_sha256` in `config.json` — and refuses having made zero `gh` calls.
- **discard** runs under the publishers' own `flock -w 30 -x`; refuses if `promotion.state`/`coordinator.phase` exists, or if the target deployment moved past the `{"op":"test","path":"/metadata/resourceVersion"}` precondition in `deployment-patch.json` (target read from the transaction's own `live-deployment.json`, not a hard-coded role map). The lock is released between probe and act, so the act script re-asserts both state files and exits 9 on a race, reported as REFUSED.
- **sweep** deletes only `Failed` pods whose `status.reason` is `Evicted`/`ContainerStatusUnknown` **and** whose owner has a `Running` replica in the same namespace (or which are ownerless and older than 24 h, because the dev cluster's fixture pods have no controller).

## Found by the live dry run

`spawnSync`'s 1 MB `maxBuffer` truncated the 2.4 MB `kubectl get pods -A -o json` into a bare `ssh` exit 255 with empty stderr — indistinguishable from an unreachable host. Fixed three ways: a host-side field projection (also keeps pod specs/env out of the log files), an explicit `MAX_ADAPTER_OUTPUT`, and an `adapterFailure()` helper that surfaces `spawnSync`'s `.error` instead of only stderr.

## Verification

`bun test orchestrator/infra.test.mjs` — 56 pass, 0 fail.
`bun run lint` — 0 errors, 616 warnings (ratchet 627; this PR adds none).
`bun run check` — clean.

Refusal tests are falsifiable: deleting the `coordinator.phase` check, the `resourceVersion` compare, or the Running-sibling rule each turns 3 tests red.

Live read-only dry runs against the runner are in the ticket's Handoff comment.

## Deliberately out of scope

`~/.claude/settings.json` and `~/Develop/hdkiller/docs/orgs/linear.md` are outside this repo — the doc states the allow-rule to paste rather than editing the operator's files. No `shared/commands/**` edits, so no `update-pins`/emit churn.


hdkiller:
## Fix round 1 — cold review ESCALATE addressed

Head `34d14cea`. `bun test orchestrator/infra.test.mjs` **83 pass / 0 fail** (was 56 tests, now 81 `test()` blocks — 25 new). `bun run lint` 0 errors, `bun run check` clean, prettier clean. All 14 check runs on the head SHA are green (2 skipped by design: Playwright smoke, pinned-tools smoke).

Every fix below was verified by reverting *just that fix* and confirming the named test goes red — the mapping in the "Red without the fix" column is the observed failure, not an intention.

| # | Finding | Fix | Test | Red without the fix |
| - | ------- | --- | ---- | ------------------- |
| 1 | `--candidate-role` shell injection: `; touch /tmp/PWNED #` reached the bash act script | `CANDIDATE_ROLE_RE = /^[a-z0-9_]+$/` checked in `runCli` **and** in `discardUnpreparedTransaction` before the host is touched; `discardActScript` asserts the shape and single-quotes the interpolation | `--candidate-role never reaches the runner's shell` (4 tests: CLI exit 3 with 0 ssh calls, verb-level refusal with 0 ssh calls, builder throws, bare flag refused) | 4 fail |
| 2 | `--yes --dry-run` mutated — `runCli` never read `flags["dry-run"]` | contradictory pair refuses with `EXIT.CANNOT_EVALUATE` before the logger and before any adapter | `--yes --dry-run is contradictory, not permissive` (one per verb asserting `ssh`/`gh`/`kubectlPods` all zero calls, plus message + "neither flag alone is affected") | 4 fail |
| 3 | host-read `binding` unquoted on the `sudo` line | `BINDING_RE = /^sha256:[0-9a-f]{64}$/` in `evaluateDiscard` → exit 3; quoted as `--recover '<binding>'` | `host-read values are inputs too` (metacharacter binding is exit 3; the verb stops after the probe, so the act script is never built; legitimate binding is quoted) | 3 fail |
| 4 | `--repos` took any slug | `DRIVER_DIGEST_REPOS` allow-list (`legalease`/`watt-mind/legalease`, `lawz`/`watt-mind/lawz`), resolved before the ssh probe | `--repos is an allow-list` (resolution table; off-list repo is exit 3 with 0 ssh and 0 gh calls) | 2 fail |
| 5 | `--host` unvalidated; a leading dash is consumed by ssh as an option | `HOST_RE` checked in `runCli`, and again in `defaultSsh` — the chokepoint every verb goes through, so a bad destination reads as a failed adapter rather than an argv entry | `--host is an ssh destination, not an option` (regex table; `-oProxyCommand=…` is exit 3 with 0 adapter calls; `defaultSsh` refuses without spawning) | 2 fail |
| 6 | post-removal `set -e` window turned a completed discard into "CANNOT EVALUATE … actions: 0" | errexit is never enabled; each pre-removal step carries its own `\|\| exit`; after the removal the fsync and journal tail are `if/else` with the status reported as a `WARNING` on stdout and `exit 0` | `the act script cannot lose a completed discard` — one structural, one that **runs the real epilogue under `bash -euo pipefail`** with `python3` and `sudo` shimmed to exit 7 and asserts exit 0 + both warnings | 2 fail |
| 7 | `resourceVersion` was only checked in the probe, whose lock is released before the act | the act script re-reads it with one `kubectl get` under the re-taken lock and compares before the `rm`; moved → exit 9 (REFUSED), unreadable → exit 10 (CANNOT EVALUATE) | `the act script re-reads the live resourceVersion under the re-taken lock, before the rm` (asserts ordering: flock < kubectl < `rm -rf`), plus `a deployment that moved between probe and act reports REFUSED` | 1 fail |
| 8 | `sweepEvictedPods` returned OK when every delete failed | `selected.length > 0 && actions === 0` → `EXIT.CANNOT_EVALUATE` | `a sweep that deleted nothing is not a clean sweep` (and: selecting nothing at all is still a clean exit 0) | 1 fail |
| 9a | a mid-loop `gh variable set` failure did not say which repos were already rotated | the error names them and names where it stopped | `a mid-loop gh failure names the repos already rotated` | 1 fail |
| 9b | log dir created world-readable | `mkdirSync(LOG_DIR, { recursive: true, mode: 0o700 })` | `makeLogger creates it 0700` | 1 fail |
| 9c | flock cost of the dry run undocumented | documented in `docs/infra-recovery.md` under the discard verb ("The dry run is not free on the host") | docs | — |

### Docs (`docs/infra-recovery.md`)

The allow-rule section is rewritten. It now says **do not** add `Bash(factory infra *)` as a blanket rule and says why — Claude Code's `Bash` rules are prefix rules and cannot tell `sweep-evicted-pods` from `sweep-evicted-pods --yes`, so the one pattern allows the mutations too. It recommends allowing the flag-free verb invocations if the permission layer can express that, keeping the prompt on `--yes`, and no rule at all if the layer can only match prefixes. It states plainly that the value is the bounded preconditions, the printed plan and the audit log — not skipping the prompt.

The sentence claiming "with `--yes` each is confined to the preconditions above" is gone. The replacement keeps the caveat even now that the preconditions hold: without `--yes` the verbs are read-only but not effect-free (the discard probe takes the publishers' lock for up to 30 s).

A new subsection states that **`discard-unprepared-transaction`'s act path has never run against a real stranded transaction** — CLNT-3170 was cleared by hand before the verb existed — and that the first real `--yes` should be operator-watched. The module header no longer advertises the blanket rule either.

### Two things beyond the list

Both are the same class as finding 3 (host-read values in a shell command) and were fixed alongside it:

- `live-deployment.json`'s `metadata.namespace`/`name` were interpolated unquoted into `kubectl -n … get deployment …`. They are now checked against RFC 1123 (exit 3 otherwise, before `kubectl` is called) and quoted.
- `kubectl delete pod -n <ns> <name>` took cluster-read names unquoted. `selectEvictedPods` now keeps (never deletes) a pod whose namespace or name is not RFC 1123, and the command is quoted. This changed one existing assertion, from `kubectl delete pod -n office-abc husk-a` to the quoted form.

